PR #2, opened 2025-11-20
Improve build system

BjornTheProgrammer:
A `rust-toolchain.toml` makes it easier to build a project that depends on a nightly toolchain. Additionally, it will install any components needed and set the valid targets.

### DIFF
--- a/mkrootfs-aarch64.sh
+++ b/mkrootfs-aarch64.sh
@@ -4,15 +4,16 @@ set -euo pipefail
 IMG="test.img"
 MOUNT="/mnt/moss"
 
-sudo mkdir -p "$MOUNT"
-
-if [ ! -f "$IMG" ]; then
-    dd if=/dev/zero of="$IMG" bs=1M count=128
-    mkfs.vfat -F 32 "$IMG"
+if [ -f "$IMG" ]; then
+    echo "initrd file $IMG already exists.  Skipping."
+    exit 0
 fi
 
+dd if=/dev/zero of="$IMG" bs=1M count=128
+mkfs.vfat -F 32 "$IMG"
+
 if ! mountpoint -q "$MOUNT"; then
-    sudo mount -o loop "$IMG" "$MOUNT"
+    mount -o loop "$IMG" "$MOUNT"
 fi
 
 if [ ! -d /tmp/bash ]; then
@@ -44,5 +45,5 @@ popd
 sudo mkdir -p "$MOUNT/dev"
 
 if mountpoint -q "$MOUNT"; then
-    sudo umount "$MOUNT"
+    umount "$MOUNT"
 fi


README.md
```diff
@@ -77,9 +77,11 @@ rustup component add rust-src llvm-tools-preview
 sudo apt install qemu-system-aarch64
 ```
 
+Additionally you will need a version of the [aarch64-none-elf](https://developer.arm.com/Tools%20and%20Software/GNU%20Toolchain) toolchain installed. To do this, download the correct release of `aarch64-none-elf` onto your computer, unpack it, then export the `bin` folder to path.
+
```
hexagonal-sun, line 89: Might be worth adding that if you're on NixOS you can get this by running:

```bash
nix shell nixpkgs#pkgsCross.aarch64-embedded.stdenv.cc nixpkgs#pkgsCross.aarch64-embedded.stdenv.cc.bintools
```

That's how I do it.  I should probably create a `flake.nix` to set up any tools that people will need.

BjornTheProgrammer: Is it actually possible to not use objcopy from the aarch64 toolchain whatsoever?

hexagonal-sun: That's a good point, is that toolchain needed just for 'objcopy'?

I think you're right, though, any version of objcopy should do the trick.

BjornTheProgrammer: I don't think it is actually possible, since the std library for aarch64 still needs to be included into the final bin

BjornTheProgrammer:
Rebased and added `test.img` generation script

qemu-runner.sh
```diff
@@ -4,6 +4,8 @@ set -e
 ELF="$1"
 BIN="${ELF%.elf}.bin"
 
+sudo bash -c "./mkrootfs-aarch64.sh"
```
hexagonal-sun: I don't think running a command via `sudo` which is invoked after a `cargo build` is great from a security standpoint.

Maybe it would be better to check whether the initrd exists and, if not, print out a helpful error message suggesting to run the creation script instead? That way users aren't suprised that `sudo` is asking for a password after running `cargo run`.

mkrootfs-aarch64.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+IMG="test.img"
+MOUNT="/mnt/moss"
+
+sudo mkdir -p "$MOUNT"
+
+if [ ! -f "$IMG" ]; then
```
hexagonal-sun: I like the idea of this not doing anything if the script exists.

Except, I'd flip the logic around:

```bash
#!/usr/bin/env bash
set -euo pipefail

IMG="test.img"
MOUNT="/mnt/moss"

if [ -f "$IMG" ]; then
    echo "initrd file $IMG already exists.  Skipping."
    exit 0
fi

[...]
```

qemu-runner.sh
BjornTheProgrammer: Great point

hardliner66: especially since the commands inside mkrootfs-aarch64.sh already use sudo. Calling the whole script with sudo runs more commands as root than necessary.

Also, it will create the image as root, which means you need to use sudo to delete it again.

Edit:
The line that creates the `dev` subdirectory needs to be run with sudo as well.
Same goes for `bin`.

hardliner66:
I'm currently testing this on WSL and I had to install dosfstools to get the script to run, because mkfs.vfat wasn't installed by default.

It would probably be a good idea to either check for the tool and install it or at least add it to the pre-requesites section:
`sudo apt install dosfstools`

hardliner66:
I also had to install `gcc-aarch64-linux-gnu` because bash couldn't be built otherwise.

hexagonal-sun:
@hardliner66 Great, thatnks for the info.  Did you get it building and working under WSL out of interest?

hardliner66:
@hexagonal-sun I'm still working on it. I commented everything that tripped me up so far in here, to make sure I don't forget reporting it. I'll also add a diff later for all the changes I made to get it to work, like missing sudo calls in the script to build the filesystem

hardliner66:
@BjornTheProgrammer I created a pull request for you with all the changes I made to get everything to work.

@hexagonal-sun I got it to start in qemu, but I'm getting a kernel panic:
image

I'm not skilled in kernel dev enough to know whats wrong.

hexagonal-sun:
@hardliner66 That's a syscall that moss hasn't yet implemented; specifically, [set_robust_list](https://linux.die.net/man/2/set_robust_list).  I suspect I don't see this issue since I compile bash against [musl](https://www.musl-libc.org/) libc, rather than gnu libc.  If you can get hold of a aarch64 musl compiler and use that instead, it should work.

I'll look to use your script to create a gilbc based initramfs and implement the missing syscalls.

Meanwhile, here is the initrd that I'm using to boot with, this should get you going hopefully.

[test.img.gz](https://github.com/user-attachments/files/23677462/test.img.gz)

hardliner66:
I'll try to get musl to work when I have the time. We should probably add a variable that can be overwritten with the target the user wants, to make it easier to support both. I'll also try to get busybox to build instead of relying on a pre built version.

hardliner66:
While trying to get everything to work, I noticed that the git repository for bash is an unofficial mirror. I tried using the official repo, but the gnu servers were unreachable, so I tried building busybox in the meantime, which could also act as a shell. Sadly, busybox didn't want to compile and I haven't figured out how to fix that yet.

I also tried building toybox instead of busybox which works, but also crashes at runtime:
image

You can check out the toybox branch here: https://github.com/hardliner66/moss/tree/toybox
It still contains the code for bash I used to cross compile for both musl and gnu, using [dockcross](https://github.com/dockcross/dockcross). I used dockcross just because my version of ubuntu doesn't provide the cross compiler thats needed and I was too lazy to build it myself.

BjornTheProgrammer:
I've merged your PR with mine

hexagonal-sun:
This is looking good!  I've managed to build the initrd from the script and replicate the errors @hardliner66 has been seeing.  There were a bunch of syscalls missing, which I've now implemented and pushed to master.  It should now boot and run hopefully!

hexagonal-sun:
Happy with the code, merging.  Thanks all!

PR #7, opened 2025-11-22
make filesystem without sudo

hardliner66:
I updated the scripts to work without sudo. I also changed that bash and busybox will be built/downloaded into `./build` and split the script into multiple files.

I also reverted the boot options back to use bash, because `/bin/sh` doesn't exist. In addition, I added a commented out call to kmain that runs `busybox sh`. If you want to use the busybox version, we could get rid of bash completely, but thats a different decision. I also changed the panic message for automount to include the file path that the code tried to mount, so you see what path couldn't be mounted.

When trying to run with `bash`, I still get the unhandled syscall panic for syscall 0x63. When trying to run with `busybox sh` it fails with:
image

### DIFF
--- a/src/arch/arm64/boot/mod.rs
+++ b/src/arch/arm64/boot/mod.rs
@@ -9,7 +9,10 @@ use crate::{
         fdt_prober::{probe_for_fdt_devices, set_fdt_va},
         init::run_initcalls,
     },
-    interrupts::{cpu_messenger::cpu_messenger_init, get_interrupt_root},
+    interrupts::{
+        cpu_messenger::{Message, cpu_messenger_init, message_cpu},
+        get_interrupt_root,
+    },
     kmain,
     memory::{INITAL_ALLOCATOR, PAGE_ALLOC},
     sched::{sched_init_secondary, uspc_ret::dispatch_userspace_task},
@@ -133,6 +136,9 @@ fn arch_init_stage2(frame: *mut ExceptionState) -> *mut ExceptionState {
     // );
 
     boot_secondaries();
+
+    // Prove that we can send IPIs through the messenger.
+    let _ = message_cpu(1, Message::Ping(ArchImpl::id() as _));
 
     frame
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,35 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  workflow_dispatch:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # Use the toolchain from rust-toolchain.toml
+      - name: Install Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+
+      - name: Verify installed toolchain and targets
+        run: |
+          rustc --version
+          rustup show
+          rustup target list --installed
+
+      - name: Check formatting
+        run: cargo fmt --all -- --check
+
+      - name: Run clippy
+        run: cargo clippy -- -D warnings
+
+      - name: Run unit tests for libkernel
+        run: cargo test -p libkernel --target x86_64-unknown-linux-gnu --all-features
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
 channel = "nightly"
-components = [ "rust-src", "llvm-tools-preview" ]
+components = [ "rust-src", "llvm-tools-preview", "rustfmt", "clippy" ]
 targets = [ "aarch64-unknown-none-softfloat" ]
--- a/src/arch/arm64/boot/secondary.rs
+++ b/src/arch/arm64/boot/secondary.rs
@@ -104,7 +104,7 @@ fn find_enable_method(node: &fdt_parser::Node<'static>) -> Result<EntryMethod> {
 fn prepare_for_secondary_entry() -> Result<(PA, PA)> {
     static mut SECONDARY_BOOT_CTX: MaybeUninit<SecondaryBootInfo> = MaybeUninit::uninit();
 
-    let entry_fn = kfunc_pa!(do_secondary_start as usize);
+    let entry_fn = kfunc_pa!(do_secondary_start as *const () as usize);
     let boot_stack = ksym_pa!(__boot_stack);
     let ctx = ksym_pa!(SECONDARY_BOOT_CTX);
 
--- a/src/drivers/timer/mod.rs
+++ b/src/drivers/timer/mod.rs
@@ -50,6 +50,7 @@ impl PartialEq for WakeupEvent {
 
 impl Eq for WakeupEvent {}
 
+#[allow(clippy::non_canonical_partial_ord_impl)]
 impl PartialOrd for WakeupEvent {
     fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
         Some(self.when.cmp(&other.when).reverse())
@@ -93,6 +94,7 @@ impl Sub<Instant> for Instant {
     }
 }
 
+#[allow(clippy::non_canonical_partial_ord_impl)]
 impl PartialOrd for Instant {
     fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
         self.ticks.partial_cmp(&other.ticks)
--- a/src/fs/syscalls/chdir.rs
+++ b/src/fs/syscalls/chdir.rs
@@ -1,7 +1,15 @@
-use crate::{fs::VFS, memory::uaccess::{copy_to_user_slice, cstr::UserCStr}, sched::current_task};
-use core::{ffi::c_char, str::FromStr};
+use crate::{
+    fs::VFS,
+    memory::uaccess::{copy_to_user_slice, cstr::UserCStr},
+    sched::current_task,
+};
 use alloc::{ffi::CString, string::ToString};
-use libkernel::{error::{KernelError, Result}, fs::path::Path, memory::address::{TUA, UA}};
+use core::{ffi::c_char, str::FromStr};
+use libkernel::{
+    error::{KernelError, Result},
+    fs::path::Path,
+    memory::address::{TUA, UA},
+};
 
 pub async fn sys_getcwd(buf: UA, len: usize) -> Result<usize> {
     let task = current_task();
--- a/src/interrupts/cpu_messenger.rs
+++ b/src/interrupts/cpu_messenger.rs
@@ -4,8 +4,6 @@ use crate::{
     arch::ArchImpl,
     drivers::Driver,
     kernel::kpipe::KBuf,
-    process::Task,
-    sched,
     sync::{OnceLock, SpinLock},
 };
 use alloc::{sync::Arc, vec::Vec};
@@ -13,7 +11,7 @@ use libkernel::{
     CpuOps,
     error::{KernelError, Result},
 };
-use log::warn;
+use log::{info, warn};
 
 use super::{
     ClaimedInterrupt, InterruptConfig, InterruptDescriptor, InterruptHandler, get_interrupt_root,
@@ -21,13 +19,14 @@ use super::{
 
 #[derive(Clone)]
 pub enum Message {
-    Reschedule,
-    PutTask(Arc<Task>),
+    // Reschedule,
+    // PutTask(Arc<Task>),
+    Ping(u32),
 }
 
 struct CpuMessenger {
     mailboxes: SpinLock<Vec<KBuf<Message>>>,
-    irq: ClaimedInterrupt,
+    _irq: ClaimedInterrupt,
 }
 
 impl Driver for CpuMessenger {
@@ -48,8 +47,11 @@ impl InterruptHandler for CpuMessenger {
             .try_pop();
 
         match message {
-            Some(Message::Reschedule) => return, // We reschedule when returning from an IRQ.
-            Some(Message::PutTask(task)) => sched::insert_task(task),
+            // Some(Message::Reschedule) => return, // We reschedule when returning from an IRQ.
+            // Some(Message::PutTask(task)) => sched::insert_task(task),
+            Some(Message::Ping(cpu_id)) => {
+                info!("CPU {} recieved ping from CPU {}", ArchImpl::id(), cpu_id)
+            }
             None => warn!("Spurious CPU IPI"),
         }
     }
@@ -74,7 +76,7 @@ pub fn cpu_messenger_init(num_cpus: usize) {
 
                 CpuMessenger {
                     mailboxes: SpinLock::new(mailboxes),
-                    irq,
+                    _irq: irq,
                 }
             },
         )
--- a/src/main.rs
+++ b/src/main.rs
@@ -44,7 +44,7 @@ fn on_panic(info: &PanicInfo) -> ! {
 
     let panic_msg = info.message();
 
-    error!("Kernel Panic: {panic_msg}");
+    error!("Kernel panic - not syncing: {panic_msg}");
 
     ArchImpl::halt();
 }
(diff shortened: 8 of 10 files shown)


src/arch/arm64/boot/mod.rs
```diff
@@ -123,10 +123,15 @@ fn arch_init_stage2(frame: *mut ExceptionState) -> *mut ExceptionState {
     cpu_messenger_init(cpu_count());
 
     kmain(
-        "--init=/bin/sh --rootfs=fat32fs --automount=/dev,devfs".to_string(),
+        "--init=/bin/bash --rootfs=fat32fs --automount=/dev,devfs".to_string(),
         frame,
     );
 
+    // kmain(
+    //     "--init='/bin/busybox sh' --rootfs=fat32fs --automount=/dev,devfs".to_string(),
```
hexagonal-sun, line 134: Might be worth raising an issue to get the kernel params from the dtb.  When you pass params to the kernel via u-boot or qemu, the bootloader embeds those arguments into the `chosen` fdt node.

hexagonal-sun:
So, I've been trying to get these scripts running on my system.  I'm having some issues.

Firstly, when I get an aarch64 libc compiler from my distro (nixos), the triplet is actually `aarch64-unknown-linux-gnu-gcc`, so I've had to modify the configure line in `build-bash.sh`.  However after modifying that I'm facing compilation issues:

```
	  ***********************************************************
	  *                                                         *
	  * GNU bash, version 5.3.0(2)-release (aarch64-unknown-linux-gnu)
	  *                                                         *
	  ***********************************************************

making lib/intl/libintl.a in ./lib/intl
make[1]: Entering directory '/home/matthew/Development/moss/build/bash/lib/intl'
aarch64-unknown-linux-gnu-gcc -c -DLOCALEDIR=\"/usr/local/share/locale\" -DLOCALE_ALIAS_PATH=\"/usr/local/share/locale\" -DLIBDIR=\"/usr/local/lib\" -DIN_LIBINTL -DBUILDING_LIBINTL -DHAVE_CONFIG_H -I. -I. -I/home/matthew/Development/moss/build/bash -I../..  -g -O2 -DLOCALEDIR=\"/usr/local/share/locale\" -DLOCALE_ALIAS_PATH=\"/usr/local/share/locale\" -DLIBDIR=\"/usr/local/lib\" -DIN_LIBINTL -DBUILDING_LIBINTL -DHAVE_CONFIG_H -DHAVE_CONFIG_H -DSHELL  l10nflist.c
l10nflist.c:62:14: error: static declaration of ‘stpcpy’ follows non-static declaration
   62 | static char *stpcpy (char *dest, const char *src);
      |              ^~~~~~
In file included from l10nflist.c:28:
/nix/store/5xbx6g3r8igyrlb2y582yi9qfbxjmjik-aarch64-unknown-linux-gnu-gcc-14.3.0/aarch64-unknown-linux-gnu/sys-include/string.h:491:14: note: previous declaration of ‘stpcpy’ with type ‘char *(char * restrict,  const char * restrict)’
  491 | extern char *stpcpy (char *__restrict __dest, const char *__restrict __src)
      |              ^~~~~~
make[1]: *** [Makefile:132: l10nflist.o] Error 1
make[1]: Leaving directory '/home/matthew/Development/moss/build/bash/lib/intl'
make: *** [Makefile:768: lib/intl/libintl.a] Error 1
```

I wonder whether using [dockcross](https://github.com/dockcross/dockcross) might help here?

Also some other quick points:

* The root of the project is getting a bit busy.  Is it possible to move the scripts into the `etc/` directory?
* When using `busybox sh` as the init command: I haven't implemented passing arguments to init yet. So the kernel is actually looking for a file called `busybox sh` and executing that, rather than passing `sh` as an argument.  Feel free to look into that if you'd like!
* syscall 0x63: I still need to implement that, but I need to understand what the kernel's role is regarding robust futexes.

hardliner66:
I had the same idea about the scripts, but didn't do it yet.

I do have a possible solution for the bash situation. How about we drop bash completely for now and just use busybox. busybox can be symlinked to the name of the tool it should run, which makes it act like that tool. FAT doesn't support symlinks, but it seems like renaming busybox works as well, so we could just rename busybox to sh and copy that to the image.

If bash is wanted, it might be a good idea to properly think that through and integrated it in a way that works for all platforms.

hexagonal-sun:
> I had the same idea about the scripts, but didn't do it yet.

No problem.


> I do have a possible solution for the bash situation. How about we drop bash completely for now and just use busybox. busybox can be symlinked to the name of the tool it should run, which makes it act like that tool. FAT doesn't support symlinks, but it seems like renaming busybox works as well, so we could just rename busybox to sh and copy that to the image.

Yes, I thnk that's a good idea.  That simplifies the process; we only need to get one userspace binary building in that case.  Have you got busybox building or are we still using the prebuilt binary?

> If bash is wanted, it might be a good idea to properly think that through and integrated it in a way that works for all platforms.

I think if busybox's inbuilt shell give us enough functionality to test various bits (pipes, redirects, launching process, etc), then I think we can drop bash.

hardliner66:
I still use the pre-built binary. I first want to get the rest of the setup to work, then I'll tackle busybox.

I came to a realization though. It seems that no matter what I try, there always seems to be a problem with the file system, unless I mount the image using a proper loop device. I also tried using docker, but then I need to run docker with --privileged, which might also pose a security risk.

So I think the best way forward is probably to split the scripts, kinda like I've done here, and let the user run them manually. This way they can call the script to make the rootfs with sudo explicitly and there are no surprises.

The whole thing seems a tad more complicated than anticipated, so I'm gonna create an issue to discuss how to move forward.

IAmTheNerdNextDoor:
Hello!

I've been trying to build moss from this PR on macOS 26.1. While I've gotten everything to build (including bash, I'd be happy to provide instructions on supporting macOS builds), moss just kernel panics on boot.

Here are my logs:
```
nerdnextdoor@Kawa moss % rm test.img                                    
nerdnextdoor@Kawa moss % cargo run --release                            
warning: function `set_icc_sgi1r_el1` is never used
  --> src/drivers/interrupts/arm_gic_v3.rs:53:4
   |
53 | fn set_icc_sgi1r_el1(value: u64) {
   |    ^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: method `raise_ipi` is never used
  --> src/interrupts/mod.rs:54:8
   |
44 | pub trait InterruptController: Send + Sync {
   |           ------------------- method in this trait
...
54 |     fn raise_ipi(&mut self, target_cpu_id: usize);
   |        ^^^^^^^^^

warning: method `raise_ipi` is never used
   --> src/interrupts/mod.rs:170:12
    |
 81 | impl InterruptManager {
    | --------------------- method in this implementation
...
170 |     pub fn raise_ipi(&self, cpu: usize) {
    |            ^^^^^^^^^

warning: variants `Reschedule` and `PutTask` are never constructed
  --> src/interrupts/cpu_messenger.rs:24:5
   |
23 | pub enum Message {
   |          ------- variants in this enum
24 |     Reschedule,
   |     ^^^^^^^^^^
25 |     PutTask(Arc<Task>),
   |     ^^^^^^^
   |
   = note: `Message` has a derived impl for the trait `Clone`, but this is intentionally ignored during dead code analysis

warning: field `irq` is never read
  --> src/interrupts/cpu_messenger.rs:30:5
   |
28 | struct CpuMessenger {
   |        ------------ field in this struct
29 |     mailboxes: SpinLock<Vec<KBuf<Message>>>,
30 |     irq: ClaimedInterrupt,
   |     ^^^

warning: function `message_cpu` is never used
  --> src/interrupts/cpu_messenger.rs:88:8
   |
88 | pub fn message_cpu(cpu_id: usize, message: Message) -> Result<()> {
   |        ^^^^^^^^^^^

warning: method `with_umask` is never used
  --> src/process/thread_group/builder.rs:41:12
   |
23 | impl ThreadGroupBuilder {
   | ----------------------- method in this implementation
...
41 |     pub fn with_umask(mut self, umask: u32) -> Self {
   |            ^^^^^^^^^^

warning: direct cast of function item into an integer
   --> src/arch/arm64/boot/secondary.rs:107:49
    |
107 |     let entry_fn = kfunc_pa!(do_secondary_start as usize);
    |                                                 ^^^^^^^^
    |
    = note: `#[warn(function_casts_as_integer)]` on by default
help: first cast to a pointer `as *const ()`
    |
107 |     let entry_fn = kfunc_pa!(do_secondary_start as *const () as usize);
    |                                                 ++++++++++++

warning: `moss` (bin "moss") generated 8 warnings (run `cargo fix --bin "moss" -p moss` to apply 1 suggestion)
    Finished `release` profile [optimized] target(s) in 0.04s
warning: the following packages contain code that will be rejected by a future version of Rust: memchr v2.7.6
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running `/Users/nerdnextdoor/Projects/moss/./qemu-runner.sh target/aarch64-unknown-none-softfloat/release/moss`
~/Projects/moss ~/Projects/moss
~/Projects/moss/build ~/Projects/moss
~/Projects/moss
128+0 records in
128+0 records out
134217728 bytes transferred in 0.032645 secs (4111432930 bytes/sec)
mkfs.fat 4.2 (2021-01-31)
/dev/disk4          	                               	/Users/nerdnextdoor/Projects/moss/build/mount
"disk4" ejected.
[    0.000000] arch::arm64::boot::memory: Adding memory region from FDT Px40000000 (0x80000000 bytes)
[    0.000000] arch::arm64::boot::memory: Reserving kernel text Px40080000 - Px41117000
[    0.000000] arch::arm64::boot::memory: Reserving FDT Px50000000 (0x100000 bytes)
[    0.000000] arch::arm64::boot::memory: Reserving initrd 0x48000000 - 0x50000000
[    0.000000] libkernel::memory::page_alloc: Buddy allocator initialized. Managing 524288 pages, 463872 free.
[    0.000000] drivers::fs::dev: devfs initialized
[    0.000000] drivers::interrupts::arm_gic_v3: ARM GICv3 found: dist @ <0x8000000, 0x10000>, rdist @ <0x80a0000, 0xf60000> (stride=0x20000)
[    0.000000] drivers::interrupts::arm_gic_v3: GICv3: Redistributor for core 0 (MPIDR=0x80000000) is awake.
[    0.000000] drivers::interrupts::arm_gic_v3: GICv3: CPU interface for core 0 enabled.
[    0.000000] interrupts: Setting device intc@8000000 as the interrupt root
[    0.000000] interrupts: Device timer claimed interrupt: Ppi(14)
[    0.002716] interrupts: Device pl011@9000000 claimed interrupt: Spi(1)
[    0.006084] libkernel::sync::per_cpu: Setup 4 per_cpu objects.
[    0.007556] interrupts: Device CPU Messenger claimed interrupt: Ipi(0)
[    0.088584] moss: Kernel Panic: Could not find automount path: /dev. Filesystem operation failed: The path or file was not found
```

I've only changed one script to mount and unmount correctly on macOS.

mkrootfs-aarch64.sh:
```bash
#!/usr/bin/env bash
set -euo pipefail

img="$base/test.img"
mount="$base/build/mount"

if [ -f "$img" ]; then
    echo "initrd file $img already exists.  Skipping."
    exit 0
fi

mkdir -p "$mount"

dd if=/dev/zero of="$img" bs=1M count=128
mkfs.vfat -F 32 "$img"

if ! mount | grep -q "$mount"; then
    hdiutil attach -mountpoint "$mount" "$img"
fi

mkdir -p "$mount/bin"
mkdir -p "$mount/dev"

cp build/bash/bash "$mount/bin"

# busybox -- I couldn't get this to build.  I ended up restoring to a third-party static binary which isn't ideal but it get's things running.
cp build/busybox-aarch64-linux-gnu "$mount/bin"

mounted=$(mount | grep "on $mount " | awk '{print $1}')
if [ -n "$mounted" ]; then
    hdiutil detach "$mounted"
fi
```

If anyone could help me with this, it would be greatly appreciated as I really want to try out moss properly. Thanks!

hexagonal-sun:
@IAmTheNerdNextDoor that error looks as though moss isn't seeing a directory at `/dev/` on your image.  Can you mount it on macOS and ensure that there is a directory at `dev`?

IAmTheNerdNextDoor:
@hexagonal-sun I can confirm that /dev does in fact exist on the image. I had checked this prior but forgot to mention it.

``` 
    Running `/Users/nerdnextdoor/Projects/moss/./qemu-runner.sh target/aarch64-unknown-none-softfloat/release/moss`
~/Projects/moss ~/Projects/moss
~/Projects/moss/build ~/Projects/moss
~/Projects/moss
128+0 records in
128+0 records out
134217728 bytes transferred in 0.025076 secs (5352437709 bytes/sec)
mkfs.fat 4.2 (2021-01-31)
/dev/disk4          	                               	/Users/nerdnextdoor/Projects/moss/build/mount
"disk4" ejected.
[    0.000000] arch::arm64::boot::memory: Adding memory region from FDT Px40000000 (0x80000000 bytes)
[    0.000000] arch::arm64::boot::memory: Reserving kernel text Px40080000 - Px41117000
[    0.000000] arch::arm64::boot::memory: Reserving FDT Px50000000 (0x100000 bytes)
[    0.000000] arch::arm64::boot::memory: Reserving initrd 0x48000000 - 0x50000000
[    0.000000] libkernel::memory::page_alloc: Buddy allocator initialized. Managing 524288 pages, 463872 free.
[    0.000000] drivers::fs::dev: devfs initialized
[    0.000000] drivers::interrupts::arm_gic_v3: ARM GICv3 found: dist @ <0x8000000, 0x10000>, rdist @ <0x80a0000, 0xf60000> (stride=0x20000)
[    0.000000] drivers::interrupts::arm_gic_v3: GICv3: Redistributor for core 0 (MPIDR=0x80000000) is awake.
[    0.000000] drivers::interrupts::arm_gic_v3: GICv3: CPU interface for core 0 enabled.
[    0.000000] interrupts: Setting device intc@8000000 as the interrupt root
[    0.000000] interrupts: Device timer claimed interrupt: Ppi(14)
[    0.002722] interrupts: Device pl011@9000000 claimed interrupt: Spi(1)
[    0.006116] libkernel::sync::per_cpu: Setup 4 per_cpu objects.
[    0.007586] interrupts: Device CPU Messenger claimed interrupt: Ipi(0)
[    0.087876] moss: Kernel Panic: Could not find automount path: /dev. Filesystem operation failed: The path or file was not found
QEMU: Terminated
nerdnextdoor@Kawa moss % hdiutil attach -mountpoint build/mount test.img
/dev/disk4          	                               	/Users/nerdnextdoor/Projects/moss/build/mount
nerdnextdoor@Kawa moss % ls build/mount                                 
bin	dev
nerdnextdoor@Kawa moss % ls build/mount/*
build/mount/bin:
bash				busybox-aarch64-linux-gnu

build/mount/dev:
```

hardliner66:
@IAmTheNerdNextDoor the problem is, that mounting the image with anything other than `sudo mount -o loop` results in an image that has the dev directory, but where it isn't recognized by the kernel. I haven't figured out what exactly the difference is.

hexagonal-sun:
@IAmTheNerdNextDoor that looks like it could be a bug in the fat driver. Could you raise an issue for it an attach a zipped/gripped version of your ram disk image please?

IAmTheNerdNextDoor:
@hardliner66 I would hazard a guess to writes not being done properly? Either that or something else. Keep in mind, I haven't looked into this too deeply and this is where I lose all status of knowing why it happens.

This is really weird behaviour. I would also probably say that maybe the loop system handles FAT differently to other systems?

@hexagonal-sun I'll do that just now.

IAmTheNerdNextDoor:
Done, issue up at #11.

IAmTheNerdNextDoor:
Speaking of this PR, we should probably switch mount methods (hdiutil on Darwin, mount on Linux and some other Unix systems) depending on what system is used. I suggest a `uname` check maybe?

hardliner66:
@IAmTheNerdNextDoor I already have a branch that uses the proper mount command, because it seems there is no easy way to create the image without it. There might be a way to "fix" the filesystem for this version, but for that someone would need to dig deeper into the differences between the two images. The other branch uses only busybox, but that brings other problems, which is why I didn't make the PR yet.

There's some discussion in #10 on how to move forward regarding what shell to focus on. I want to wait until thats decided before making yet another PR.

hardliner66:
Closed in favor of #13

PR #9, opened 2025-11-22
main: change panic message to match Linux

Cloudwalk9:
It is imperative for true Linux compatibility and for all of its quirkiness, that the kernel duly notify the user on fatal error conditions within the kernel that it is most definitely NOT SYNCING.


hexagonal-sun:
Indeed, when the kernel panics it most certainly doesn't sync!

PR #14, opened 2025-11-24
Add markdown file with all syscalls

hardliner66:
I used the json export from https://syscalls.mebeim.net/?table=arm64/64/aarch64/latest and wrote a small helper tool to format it to markdown. This PR adds that markdown to `etc`.

I also added an `Implemented` column, so that it can be tracked what exists and what doesn't. Some of those entries use `dummy` in the implemented column, which means there is a handler for that syscall, but its just a dummy implementation.


IAmTheNerdNextDoor:
This could be very useful! I think this should be added.

hexagonal-sun:
This is really nice! Thank you for adding this. One thing to note, the syscall numbers are specific per architecture, so I'd either store this file in src/arch/arm64 or rename to something like etc/aarch64-syscalls.md?

hardliner66:
I renamed it for now. In the future it should probably be moved into src/arch/arm64.

PR #15, opened 2025-11-24
Fix link in readme

hardliner66:
I forgot to change the link when renaming the file. This PR fixes that.


hexagonal-sun:
Ah nice catch!

PR #16, opened 2025-11-24
libkernel: fs: fat32: make lookups case-insensitive

hexagonal-sun:
Fat32 is a case-insensitive filesystem, therefore when looking up a path in a directory entry, convert all strings to lowercase before comparison.


hexagonal-sun:
@hardliner66 @BjornTheProgrammer Are you happy with this?

BjornTheProgrammer:
Yeah. It makes sense to do that check. Just something to keep in mind is that fat32 still preserves the casing.

hardliner66:
I agree with @BjornTheProgrammer on this

BjornTheProgrammer (approved):

PR #17, opened 2025-11-25
Libc program bringup fixes

hexagonal-sun:
Changes for supporting libc compiled static binaries. This fixes issues with `/bin/bash` in the initrd posted in #11.

@hardliner66 @IAmTheNerdNextDoor if you don't mind looking through and giving the changes a quick review and making sure this branch does start to fix various issues encountered with your libc binaries that would be great!

PR #21, opened 2025-11-25
libkernel: mem: use brk_addr instead

IAmTheNerdNextDoor:
To fix the broken test in pull request #17.

### DIFF
--- a/libkernel/src/memory/proc_vm/mod.rs
+++ b/libkernel/src/memory/proc_vm/mod.rs
@@ -193,7 +193,7 @@ mod tests {
 
         // The new break should be page-aligned
         let expected_brk_end = brk_addr.align_up(PAGE_SIZE);
-        assert_eq!(new_brk, expected_brk_end);
+        assert_eq!(new_brk, brk_addr);
         assert_eq!(vm.current_brk(), expected_brk_end);
         assert_eq!(vm.brk.size(), PAGE_SIZE);
 


hexagonal-sun:
Thanks for this, however the commit in the PR  returns non-page aligned addresses from brk().  I thought Linux aligned up all brk()  calls to the page boundary, but it doesn't; it returns a non-aligned address.  See:

```c
#define _GNU_SOURCE
#include <unistd.h>
#include <sys/syscall.h>
#include <stdio.h>
#include <stdint.h>

int main(void) {
    // Get current program break
    void *cur = (void *) syscall(SYS_brk, 0);
    printf("Current brk: %p\n", cur);

    // Add 123 bytes to make it obviously non-aligned
    void *non_aligned = (void *) ((uintptr_t)cur + 123);
    printf("Requesting non-aligned brk: %p\n", non_aligned);

    // Request new brk
    void *ret = (void *) syscall(SYS_brk, non_aligned);
    printf("Returned brk: %p\n", ret);

    return 0;
}
```

The output on my machine:

```
Current brk: 0x55d90b99b000
Requesting non-aligned brk: 0x55d90b99b07b
Returned brk: 0x55d90b99b07b
```

In this case, the unit tests need fixing rather than returning the non-aligned value.

IAmTheNerdNextDoor:
Ah, I see. I could probably close this PR and make a different PR to fix this unless you plan to already?

hexagonal-sun:
It's okay, just force push to this PR.  You just need to change the the second parameter at `libkernel/src/memory/proc_vm/mod.rs:196` to be `brk_addr` (the unaligned address).

IAmTheNerdNextDoor:
Right, tests work still!

hexagonal-sun:
Perfect, thanks!

PR #23, opened 2025-11-26
Introduce experimental macOS build support

IAmTheNerdNextDoor:
Tested on my M4 MacBook Air, builds fine.

Following @hardliner66's advice, they're in their own folder with a warning of being unable to guarantee it'll work for everyone.

If anyone has any objections or changes, let me know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,9 @@ Additionally you will need a version of the [aarch64-none-elf](https://developer
 #### Any OS
 To install aarch64-none-elf on any os, download the correct release of `aarch64-none-elf` onto your computer, unpack it, then export the `bin` folder to path.
 
+#### macOS
+There is experimental support for macOS in the scripts/mac-experimental folder. The scripts in there are not guaranteed to work for all macOS users and has only been tested on an M4 Apple Silicon MacBook Air.
+
 #### NixOS
 
 Run the following command
--- a/scripts/mac-experimental/README.md
+++ b/scripts/mac-experimental/README.md
@@ -1,0 +1,47 @@
+# Experimental macOS build scripts
+
+This is the folder for experimental macOS build support on Apple Silicon (we do not plan to support x86_64 macOS).
+
+These scripts are not guaranteed to work and are not actively supported by the maintainer or other contributors. 
+If you do have any issues, please mention @IAmTheNerdNextDoor in GitHub Issues.
+
+### Dependencies
+Regardless of stdlib, you will need:
+- [Rustup](https://rustup.rs) (DO NOT install Rustup from Homebrew as you will incur conflicts. If you have Rustup installed from Homebrew, uninstall it and reinstall from the Rustup site.)
+- [aarch64-none-elf](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads)
+- [Homebrew](https://brew.sh)
+- QEMU (`brew install qemu`)
+- dosfstools (`brew install dosfstools`)
+
+If building for the GNU library:
+- aarch64-unknown-linux-gnu (`brew tap messense/macos-cross-toolchains && brew install aarch64-unknown-linux-gnu`)
+
+If building for the Musl library:
+- Zig (`brew install zig`)
+
+We use Zig to cross-compile to Linux musl on macOS.
+
+When it comes to Rust, you need the nightly toolchain. To install it, run:
+`rustup default nightly`
+
+### Building
+You can start building the initrd programs using this script:
+`./scripts/mac-experimental/build-deps.sh`
+
+musl programs are built by default.
+
+If you want GNU program builds:
+`stdlib=gnu ./scripts/mac-experimental/build-deps.sh`
+
+After building, you can make the initrd with:
+`./scripts/mac-experimental/create-image.sh`
+
+The initrd should now be at `./moss.img`.
+
+You can then build and run moss with:
+`cargo run --release`
+
+If you get an error about `aarch64-unknown-none-softfloat` during building, you need to install the target:
+`rustup target install aarch64-unknown-none-softfloat`
+
+Have fun!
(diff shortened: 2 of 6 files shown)


hardliner66:
I have one minor nitpick. The linux version uses the the toolchain from musl.cc, the mac version uses zig instead, but I think we shouldn't use completely different toolchains unless its absolutely necessary.

I don't favor one toolchain over the other, but I think we should either use the zig toolchain or the toolchain from musl.cc. Otherwise we might have to deal with bugs from devs from one platform, that aren't reproducible in the other and vice-versa, due to toolchain differences.

Edit:
I just got the notification that I was mentioned here. Way after I wrote my response 😂

IAmTheNerdNextDoor:
Precompiled toolchains for macOS from musl.cc are unavailable at the moment apparently. (Check https://mac.musl.cc/)

I tried finding another native cross-compiler for musl that properly worked with moss and that didn't work. It was from Homebrew as well. 

The one I found for moss decided to compile, but not correctly somehow which caused moss to fail running musl bash. Other toolchains I found were prebuilt for Linux. Building your own toolchain is a pain in the neck on macOS.

Zig however worked for musl but not GNU. GNU had errors and issues when compiling bash using Zig. It has a working prebuilt cross-compiler though on Homebrew.

My solution here is either we keep both, or drop support for one stdlib on macOS. I vote GNU out in favour of musl as it's smaller, cleaner and less bloated (which for a kernel like moss, musl would be better over GNU). 

Using Zig as a cross-compiler for Linux musl is better too as it's installable on plenty of systems and it doesn't require a bunch of other misc stuff.

My decision to have both originally were derived from a couple hours of testing toolchain methods to require less dependencies.

hardliner66:
I think its okay to focus on musl first and add gnu later when the kernel is more complete.

I also didn't see that the mac toolchain was temporarily unavailable. I should wear my glasses more often ^^
I'll prepare a PR to switch the toolchain over to zig for linux hosts as well. Unless @hexagonal-sun doesn't want a dependency to zig, but I think it should be fine. The zig toolchain isn't completely stable yet, but every time I ran into a problem, I could just write a message to the zig irc channel and Andrew himself would take the time to help.

IAmTheNerdNextDoor:
I don't think the actual Zig part of Zig changing will be an issue. I think they won't touch cross-compile stuff.

Then again I don't know much about Zig. I've always been meaning to learn it.

hardliner66:
Okay, short update, I tried getting bash to compile with zig, but there seems to be a problem with that, so I think I'm gonna leave it as is for now.

IAmTheNerdNextDoor:
> Okay, short update, I tried getting bash to compile with zig, but there seems to be a problem with that, so I think I'm gonna leave it as is for now.

Weird. GNU bash or musl bash? musl bash has no issues, GNU does on Zig.

hardliner66:
Musl. I got it to work, but now the kernel panics 😑

IAmTheNerdNextDoor:
That's weird. Have you checked your moss.img to see if bash was built correctly?

hexagonal-sun:
@hardliner66 What's your kernel panic, is it `set_robust_list` syscall missing?  I'm hitting that at the moment when executing busybox (fix in the works).

arihant2math:
I had to use `hdiutil attach -noverify -nomount moss.img && mount -w -o rw  -t msdos  /dev/disk4 build/mount` to get macos to accept the mounting. The original `hdiutil` command would fail.

IAmTheNerdNextDoor:
What version of macOS are you on @arihant2math? I'm on 26.1.

hexagonal-sun:
One option that might help here, could we not use [mtools](http://mtools.linux.lu/mtools.html)  (I'm not sure if it's available on MacOS).  They allow you to modify a FAT filesystem image without needing to mount it.

IAmTheNerdNextDoor:
Seems like mtools is installable from Homebrew!

arihant2math:
> What version of macOS are you on @arihant2math? I'm on 26.1.

26.0

IAmTheNerdNextDoor:
> > What version of macOS are you on @arihant2math? I'm on 26.1.
> 
> 26.0

I see. That's weird... are you sure you have nothing that interferes with the mounting process or something installed?

arihant2math:
There isn't any data-loss daemon that interferes with mounting if that's what you mean.

Seems like a common problem though: https://superuser.com/questions/19426/im-unable-to-mount-a-dmg-getting-a-no-mountable-filesystems-error.

hexagonal-sun:
Are we happy to merge these changes, or is there still more to figure out with the mount issue?

IAmTheNerdNextDoor:
> Are we happy to merge these changes, or is there still more to figure out with the mount issue?

I'd be happy to merge. I'm not sure about why the issues that @arihant2math had even occurred. I don't know their exact setup specifics so I'd chalk it up to an issue on their side. If their system can't mount an image that it made I'd blame the setup.

My macOS installation is new (originally around macOS 15.6 when I got my machine in around August) and went to 26.1 in early November. I haven't modified anything relating to mounting at all and it works. The Superuser they had linked is from 16 years ago, so I doubt it's the same issue as from then.

I also think that the workaround adds more complexity than needed to mount. Maybe we could add `mtools` support or something in the future so we don't have to mount but this should work for most people for now.

arihant2math:
It's weird that the issue still occurs 14 years later, but I think it's still an issue. Anyhow merging this is probably fine; I'll keep my own patches in place.

IAmTheNerdNextDoor:
Yeah, it is weird. I think maybe at some point a fallback could be introduced if so that if mounting fails in the same way yours did, we could attempt to mount it all manually.

hexagonal-sun:
Okay - sounds like there's perhaps some more stuff to work out, but that can be addressed in separate issues.  Thanks for your help everyone.

Issue #25, opened 2025-11-27
Using a fair scheduler


arihant2math:
I took a look at the scheduling code, and thought it could be improved by either using round robin (very simple) or implementing CFS or EEVDF, which linux has used and is using respectively. I'm biased towards EEVDF, so I might take a shot at that.

PR #29, opened 2025-11-27
Implement fair scheduler (EEVDF)

arihant2math:
Original Paper: https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=805acf7726282721504c8f00575d91ebfd750564
Linux Docs: https://docs.kernel.org/scheduler/sched-eevdf.html

I haven't implemented lag, but conceivably it shouldn't be terribly difficult to implement. 

Fixes: #25


hexagonal-sun:
Nice!  Thanks for working on this.  I didn't realise that Linux was moving away from the CFS in favor of EEVDF, interesting!

PR #32, opened 2025-11-28
Properly shutdown system

arihant2math:
In case of a kernel panic, the default behavior was to just halt. I added a proper exit, so QEMU wouldn't just annoyingly hang (I had to constantly close and reopen terminals). I also add the sync syscall so that `busybox reboot` almost works (`/proc` needs to be implemented). Anyways the `reboot` syscall should partially work.


IAmTheNerdNextDoor:
You could also hit Ctrl+A, X to close QEMU from the terminal.

src/arch/arm64/mod.rs
```diff
@@ -108,6 +108,23 @@ impl Arch for Aarch64 {
         proc::idle::create_idle_task()
     }
 
+    fn power_off() -> ! {
+        // Try PSCI `SYSTEM_OFF` first (works on QEMU `-machine virt` and most
+        // real hardware that implements the PSCI interface).
+        const PSCI_SYSTEM_OFF: u32 = 0x8400_0008;
+        unsafe {
+            core::arch::asm!(
```
hexagonal-sun: There is a PSCI module in `src/arch/arm64/boot/secondary/psci.rs`. Ideally, this code would use that.

If `psci` is needed for other things than secondary bring up, feel free to move it to the `arm64` module and make it public there.

src/arch/arm64/exceptions/syscall.rs
hexagonal-sun, line 1: Please update the syscall table in `etc/syscalls_linux_aarch64.md` to show these as implemented.

hexagonal-sun (changes requested):
Looks good, if you could refactor to use the PSCI module that would be great!

PR #34, opened 2025-11-28
Implement sys_set_robust_list and sys_gettid

BjornTheProgrammer:
I'm not 100% sure whether I've implemented sys_set_robust_list correctly, but at the very least, this should help with more compatibility.


arihant2math:
I was attempting an implementation as well; `sys_set_robust_list` is similar to what I was doing.

Whenever `sys_get_robust_list` is implemented, `ptrace` checking can be skipped and we can just check if the process ids match between the target thread and the current thread.

BjornTheProgrammer:
I've rebased and updated syscalls.

BjornTheProgrammer:
> I was attempting an implementation as well; `sys_set_robust_list` is similar to what I was doing.
> 
> Whenever `sys_get_robust_list` is implemented, `ptrace` checking can be skipped and we can just check if the process ids match between the target thread and the current thread.

Perfect! I didn't bother to implement `sys_get_robust_list` since I was running into issues with `rseq`, but it would be great if you added a PR to implement that!

Issue #33, opened 2025-11-28
[bug] can't build on aarch64 linux


hpolloni:
Hi! cool project!

I'm trying to build this on an M1 Mac. I'm using an aarch64 linux running on a VM. I'm following the instructions to build the kernel and the deps. When I execute `./scripts/build-deps.sh` the script downloads a cross compiler, this probably means that the script assumes that the host machine is x86. 

```
./configure: line 4424: /home/ubuntu/moss-kernel/build/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc: cannot execute binary file: Exec format error
./configure: line 4424: /home/ubuntu/moss-kernel/build/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc: cannot execute binary file: Exec format error
./configure: line 4424: /home/ubuntu/moss-kernel/build/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc: cannot execute binary file: Exec format error
./configure: line 4424: /home/ubuntu/moss-kernel/build/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc: cannot execute binary file: Exec format error
./configure: line 4424: /home/ubuntu/moss-kernel/build/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc: cannot execute binary file: Exec format error
./configure: line 4477: /home/ubuntu/moss-kernel/build/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc: cannot execute binary file: Exec format error
configure:4525: error: in '/home/ubuntu/moss-kernel/build/bash':
configure:4527: error: C compiler cannot create executables
```

IAmTheNerdNextDoor:
Hi! I'm basically the person for macOS stuff here.

~~The toolchain that is downloaded by the normal build scripts is for host Linux systems. In PR #23, there is some experimental support for macOS that I'm trying to merge in. It works for me and uses Zig to cross-compile musl bash from macOS to Linux.~~

~~Could you try the scripts from that PR and let me know how that goes? (The scripts for that are in scripts/mac-experimental instead of the normal one.)~~

Please disregard this, I read this wrong. Hold on.

IAmTheNerdNextDoor:
Really sorry about the response I gave there. I saw exec format error and M1 Mac and my brain immediately jumped to Mac support.

[...]

PR #36, opened 2025-11-28
Fix bash build on aarch64 host

hpolloni:
See https://github.com/hexagonal-sun/moss-kernel/issues/33 -- I tested this on my Mac M1 running an ubuntu VM. Next step is to run this on a x86 host to catch any regressions


hpolloni:
Tested on an x86_64 host...works as expected.

hexagonal-sun:
LGTM. Thanks for the contribution!

Issue #37, opened 2025-11-28
Support futex


arihant2math:
This is needed to do multithreaded testing in rust, as it uses this API internally.

- [x] sys_set_robust_list #34 
- [x] ~~sys_get_robust_list~~ Needs ptrace, wait for that first.
- [x] sys_futex (this is actually two different syscalls combined into one) #57 
- [x] Bitset support
- [x] Private support
- [x] Timeout support

hexagonal-sun:
I think we're getting close on this one.  Now that #57 has been merged  and on top of #74 ,  I changed the `run_test` function on `usertest` to be:

```rust
fn run_test(test_fn: fn()) {
    thread::spawn(move || test_fn()).join().unwrap();
}
```
However, something perculiar happens, the test just end up sleeping:
```
bash-5.3# ./bin/usertest 
Running userspace tests ...
Testing sync syscall ... OK
Testing opendir syscall ... OK
bash-5.3#
```
Then what's even stranger is that the process is still running!
```
h-5.3# ./bin/busybox ls /proc
0     1     4     4     5     self
bash-5.3# ./bin/busybox cat /proc/4/status
Name:   ./bin/usertest
State:  S
Tgid:   4
FDSize: 3
Pid:    4
Threads:        0
bash-5.3# 
```

What I find strange is why `bash` becomes active when the child process is still active, I would have expected it to be still stuck in `wait4`.  The only reason it would come out of that is if it gets a signal if I remember correctly, 

My gut feeling is that the signal routing has got screwed and the spawned thread is trying to signal to it's parent thread, but that's going through to bash instead.  That would explain why bash gets awoken but `usertest` is just deadlocked.

arihant2math:
My understanding is that rust uses a futex for the join so it probably passes `CHILD_CLEARTID`, which we aren't honoring.

arihant2math:
Fixing this leads to:
```
bash-5.3# ./usertest
Running userspace tests ...
Testing sync syscall ... OK
Testing opendir syscall ... OK
Testing readdir syscall ... OK
Testing chdir syscall ... OK
Testing chroot syscall ... OK
Testing fork syscall ... OK

thread '<unnamed>' (8) panicked at usertest/src/main.rs:103:13:
open failed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' (2) panicked at usertest/src/main.rs:216:45:
called `Result::unwrap()` on an `Err` value: Any { .. }
[...]

PR #40, opened 2025-11-28
Userspace testing

arihant2math:
A simple rust program that ensures some basic syscalls work.
I think that testing filesystem operations would be way easier with `/tmp` implemented.
Testing `/dev/null` could be interesting too.

### DIFF
--- a/usertest/src/main.rs
+++ b/usertest/src/main.rs
@@ -124,4 +124,3 @@ fn main() {
     let end = std::time::Instant::now();
     println!("All tests passed in {} ms", (end - start).as_millis());
 }
-


usertest/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+fn main() {
+    println!("Running userspace tests ...");
+    let start = std::time::Instant::now();
+    // TODO: make tests their own process later
+    run_test(test_sync);
+    run_test(test_opendir);
+    run_test(test_readdir);
```
hexagonal-sun, line 120: I wonder whether there's a way to use Rust's standard test infrastructure here to run these tests.

Although, as you say - it will require threading, so we'd need #37 sorting first.

arihant2math: Also this is a little sturdier because it prevents some breakage by forking, which should help reduce bugs from causing runaway issues, given that forking works.

hexagonal-sun (approved):

hexagonal-sun:
LGTM

PR #42, opened 2025-11-30
Implement sys_kill

arihant2math:
I'm working on getting bash to work inside of bash. This is the first part, SIGSTOP needs to be implemented after this.


hexagonal-sun (approved):
LGTM

PR #46, opened 2025-12-07
Update README.md for some clarification

TheGrol:
I was having a hard time figuring out what this meant when I was going through the instructions the first time, and I believe that this would help clarify things for people who got stuck like I did. I also corrected the name for "Any OS" to clarify this is for X86-64 Linux OSes, because of the command I provided in the README.


hexagonal-sun:
Thanks for the fix

PR #47, opened 2025-12-08
Handle stop and continue `KSingalAction`s

arihant2math:


src/sched/uspc_ret.rs
```diff
@@ -193,8 +196,62 @@ pub fn dispatch_userspace_task(ctx: *mut UserCtx) {
...
+                            {
+                                parent
+                                    .child_notifiers
+                                    .child_update(process.tgid, ChildState::Stop { signal: id });
+                                parent.signals.lock_save_irq().set_pending(SigId::SIGCHLD);
+                            }
+
+                            for thr_weak in process.threads.lock_save_irq().values() {
+                                if let Some(thr) = thr_weak.upgrade() {
+                                    *thr.state.lock_save_irq() = TaskState::Sleeping;
```
hexagonal-sun: Might be worth having another state here - `Stopped`?  The reason being is that any pending wakers targetting this task will wake up a `Sleeping` task.  If we use `Stopped`, wakers won't spuriously wake the task back up.

Issue #49, opened 2025-12-09
Implementing tmpfs


arihant2math:
This would help a lot with testing things like `vi` since the main filesystem is RO.

hexagonal-sun:
WIP on the `tmpfs` branch.

PR #53, opened 2025-12-11
Tmpfs

hexagonal-sun:
`tmpfs` implementation.  We use one page as an indirect block for data.  Allows individual files of up to 2GB.   First writes now in moss!

```
bash-5.3# ./bin/busybox ls tmp/test.txt
ls: tmp/test.txt: No such file or directory
bash-5.3# ./bin/busybox vi tmp/test.txt
[ wrote some data here ]
bash-5.3# ./bin/busybox ls -l tmp/test.txt
----------    1 0        0               16 Jan  1 00:00 tmp/test.txt
bash-5.3# ./bin/busybox cat tmp/test.txt
This is a test.
bash-5.3# 
```

Missing:
  - symlinks
  -  special files (pipes, fifios, chardevs, blockdevs, etc)
  - timestamps
  - permissions seem to be a bit screwy

Also implements the `ftruncate` syscall.

Fixes: #49

### DIFF
--- a/libkernel/src/fs/filesystems/tmpfs.rs
+++ b/libkernel/src/fs/filesystems/tmpfs.rs
@@ -257,7 +257,7 @@ where
             // Free the excess blocks from the end
             while inner.allocated_blocks > new_blk_count {
                 let release_idx = inner.allocated_blocks - 1;
-                
+
                 unsafe {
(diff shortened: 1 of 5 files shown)


arihant2math (approved):
LGTM

arihant2math:
Should help a lot with the ext4 implementation, since the general VFS implementation is now there.

PR #56, opened 2025-12-14
Test read syscall in userspace

arihant2math:
Now that we have `/dev/zero/` testing that read fills the buffer is trivial.

### DIFF
--- a/usertest/src/main.rs
+++ b/usertest/src/main.rs
@@ -76,7 +76,7 @@ fn test_read() {
     print!("Testing read syscall ...");
     let file = "/dev/zero";
     let c_file = std::ffi::CString::new(file).unwrap();
-    let mut buffer = [0u8; 16];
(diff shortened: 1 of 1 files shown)


usertest/src/main.rs
```diff
@@ -72,6 +72,26 @@ fn test_fork() {
     println!(" OK");
 }
 
+fn test_read() {
+    print!("Testing read syscall ...");
+    let file = "/dev/zero";
+    let c_file = std::ffi::CString::new(file).unwrap();
+    let mut buffer = [0u8; 16];
```
hexagonal-sun: Maybe initialize this with `[1u8; 16]`? that way we know that the kernel actually overwrote this buffer with zeros?

hexagonal-sun (changes requested):

hexagonal-sun (approved):

PR #57, opened 2025-12-14
Futex

arihant2math:
WIP.

### DIFF
--- a/src/process/threading.rs
+++ b/src/process/threading.rs
@@ -28,7 +28,11 @@ impl FutexWaitQueue {
 }
 
 /// Global futex table mapping a user address to its wait queue.
-static FUTEX_TABLE: OnceLock<SpinLock<BTreeMap<usize, Arc<SpinLock<FutexWaitQueue>>>>> =
+// TODO: statically allocate an array of SpinLock<Vec<FutexWaitQueue>>.
+// Then hash into that table to find it's bucket
+// TODO: Should be physical address, not user address
+#[allow(clippy::type_complexity)]
+static FUTEX_TABLE: OnceLock<SpinLock<BTreeMap<TUA<u32>, Arc<SpinLock<FutexWaitQueue>>>>> =
     OnceLock::new();
 
 pub async fn sys_set_tid_address(_tidptr: VA) -> Result<usize> {
@@ -98,23 +102,17 @@ pub async fn sys_futex(
             let waitq_arc = {
                 let mut guard = table.lock_save_irq();
                 guard
-                    .entry(uaddr.value())
+                    .entry(uaddr)
                     .or_insert_with(|| Arc::new(SpinLock::new(FutexWaitQueue::new())))
                     .clone()
             };
 
-            // First read – may fault and therefore is done outside the lock.
             let current: u32 = copy_from_user(uaddr).await?;
             if current != val {
                 return Err(KernelError::TryAgain);
             }
 
-            // Second read – ensures the value is still what we expect *after*
-            // the wait-queue entry is visible to other threads.
-            if copy_from_user(uaddr).await? != val {
-                return Err(KernelError::TryAgain);
-            }
-
+            // TODO: When we have try_ variants of locking primitives, use them here
             wait_until(
                 waitq_arc.clone(),
                 |state| &mut state.wakers,
@@ -137,7 +135,7 @@ pub async fn sys_futex(
             let mut woke = 0;
 
             if let Some(table) = FUTEX_TABLE.get()
-                && let Some(waitq_arc) = table.lock_save_irq().get(&uaddr.value()).cloned()
+                && let Some(waitq_arc) = table.lock_save_irq().get(&uaddr).cloned()
             {
                 let mut waitq = waitq_arc.lock_save_irq();
                 for _ in 0..nr_wake {


arihant2math:
Rust mutexes panic now, so I'll want to look into that first (replicate this by running the usertest program).

arihant2math:
I'm getting
```
[    2.2178126] moss: Kernel panicked at src/arch/arm64/memory/fault.rs:110:40: SIGSEGV on process 11 InstrAbortLowerEL(AbortIss { ifsc: Ifsc(15), write: false, far: Some(140737488352240) }) PC: 7ffffffff3f0
```

but that could be some prior issue cropping up here.

src/process/threading.rs
```diff
@@ -38,3 +62,79 @@ pub async fn sys_set_robust_list(head: TUA<RobustListHead>, len: usize) -> Resul
...
+            // Fail fast if the value has already changed
+            let current: u32 = copy_from_user(uaddr).await?;
+            if current != val {
+                return Err(KernelError::TryAgain);
+            }
+
+            // Obtain (or create) the wait-queue for this futex word
+            let table = FUTEX_TABLE.get_or_init(|| SpinLock::new(BTreeMap::new()));
+            let waitq_arc = {
+                let mut guard = table.lock_save_irq();
```
hexagonal-sun, line 103: I thnk this approach could cuase a data-race, and therefore a deadlock.

I've been thinking about how we can make this non-racey. I think what we need to do is to check the value with the spinlock grabbed such that we can guarentee atomicity of the insertion of the task into the wait queue, with the integer value at the specified value.

However, we cannot check the value with the spinlock held.  One option is to use a `Mutex` instead of a spinlock so we can call `copy_from_user` while holding the lock.  Alternatively, we could come up with a new set of functions which attempt to copy from userspace but never sleep, thus allowing them to be called while a spinlock is held.  If we opt for the latter, we'd need some kind of page-in retry logic so we try to page-in the userspace address and then retry.

I'm not sure what strategy Linux uses, I need to look through to code to figure it out. Do you have any ideas?

arihant2math: I think moving the spinlock up makes sense. But now I'm wondering what happens when `Err(KernelError::TryAgain)` is returned in a normal application. Also it hangs if I run things on the same thread (creating threads triggers #66 it seems)

hexagonal-sun:
So, running usertests for me at the head of this branch seems to result in deadlock:

```
bash-5.3# ./bin/usertest 
Running userspace tests ...
Testing sync syscall ... OK
Testing opendir syscall ... OK
Testing readdir syscall ... OK
Testing chdir syscall ... OK
Testing fork syscall ... OK
Testing read syscall ... OK
Testing write syscall ... OK
Testing futex syscall ... OK
```

I'm guessing this is hanging in the `test_rust_mutex` function, we're just not seeing the `print!("Testing Rust Mutex ...");` due to buffering.  I'll have a look and see why this might be hanging.

arihant2math:
Seems to be calling sys_futex in a loop with cmd=9, and moss is returning an error that it doesn't catch, hence the hang.

arihant2math:
Seems like the flag handling needs a rework.

hexagonal-sun:
I've taken a quick look... looks like we're missing:

```rust
const FUTEX_WAIT_BITSET: i32 = 9;
const FUTEX_WAKE_BITSET: i32 = 10;
```

I've added them and just stubbed them out with their non-bitset equivalents:

```rust
    match cmd {
        FUTEX_WAIT | FUTEX_WAIT_BITSET => {
            [...]
        }
        FUTEX_WAKE | FUTEX_WAKE_BITSET => {
            [...]
        }
        _ => Err(KernelError::NotSupported),
    }
```

 When running `usertest` I now see:

```
Testing write syscall ... OK
[   10.10301407] process::threading: 0x7ffffffffc44 1 1
[   10.10302460] process::threading: 0x7ffffffffc44 0 1
Testing futex syscall ... OK
[   10.10307857] process::threading: 0x7fffff7fe050 9 2
```

I'm surprised that a rust mutex is calling futex at all -- AFAICS, it's uncontended at this point so I'm unsure why it's calling into the kernel.  Second, it's calling WAIT but there is no other active thread to call WAKE so it deadlocks, and in-fact we get the same behaviour on Linux... this might be a bug in the `usertest` program:

```
qemu-aarch64 ./build/bin/usertest
Running userspace tests ...
Testing sync syscall ... OK
Testing opendir syscall ... OK
Testing readdir syscall ... OK
Testing chdir syscall ... OK
Testing fork syscall ... OK
Testing read syscall ... OK
Testing write syscall ... OK
Testing futex syscall ... OK
```

If I run with QEMU_STRACE=1, then we get:

```
> QEMU_STRACE=1 qemu-aarch64  ./build/bin/usertest 

[...]

163682 futex(0x00007fdea9a84050,FUTEX_PRIVATE_FLAG|FUTEX_WAIT_BITSET,2,NULL,NULL,0)
```

Aha, it's hanging in the same place on Linux!  So moss is doing the 'right' thing here.

hexagonal-sun:
Ah, found it in `test_rust_mutex`:

```rust
    let mut num = mtx_clone.lock().unwrap();
    *num += 1;
    let final_count = *mutex.lock().unwrap();
```

we lock the mutex and then lock it again, `num` isn't dropped. deadlock.

arihant2math:
Great, usertest passes now.

src/process/threading.rs
```diff
@@ -1,11 +1,36 @@
...
+    fn new() -> Self {
+        Self {
+            wakers: WakerSet::new(),
+            wakeups: 0,
+        }
+    }
+}
+
+/// Global futex table mapping a user address to its wait queue.
+static FUTEX_TABLE: OnceLock<SpinLock<BTreeMap<usize, Arc<SpinLock<FutexWaitQueue>>>>> =
```
hexagonal-sun: A couple of points - probably for a later iteration, but maybe capture them as a todo:

* Ideally we'd implement this the way Linux does, statically allocate an array of `SpinLock<Vec<FutexWaitQueue>>`.  then hash into that table to find it's bucket.  That's the way Linux does it, but it's a small optimisation.
* We need to think about how we key futexs from different address spaces.  The best way of doing this is to resolve the UA (which is a virtual address) into the PA and then key off that.  I think Linux has a more sophisticated way of doing this which doesn't involve a page-table walk for every futex operation, but it's an edge-case.  Nonetheless, here's an example:

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <pthread.h>
#include <sys/mman.h>
#include <sys/syscall.h>
#include <linux/futex.h>
#include <fcntl.h>
#include <stdatomic.h>

static atomic_int *futex_a;
static atomic_int *futex_b;

static int futex_wait(atomic_int *addr, int expected)
{
    return syscall(SYS_futex, addr, FUTEX_WAIT, expected, NULL, NULL, 0);
}

static int futex_wake(atomic_int *addr, int count)
{
    return syscall(SYS_futex, addr, FUTEX_WAKE, count, NULL, NULL, 0);
}

void *waiter(void *arg)
{
    (void)arg;
    printf("Waiter waiting on VA %p\n", (void *)futex_a);

    while (atomic_load(futex_a) == 0)
        futex_wait(futex_a, 0);

    printf("Waiter woken, value=%d\n", atomic_load(futex_a));
    return NULL;
}

void *waker(void *arg)
{
    (void)arg;
    sleep(1);

    printf("Waker writing via VA %p\n", (void *)futex_b);
    atomic_store(futex_b, 1);
    futex_wake(futex_b, 1);
    return NULL;
}

int main(void)
{
    size_t page = sysconf(_SC_PAGESIZE);

    /* Create shared anonymous backing */
    int fd = memfd_create("futex-page", 0);
    if (fd < 0) {
        perror("memfd_create");
        exit(1);
    }

    if (ftruncate(fd, page) < 0) {
        perror("ftruncate");
        exit(1);
    }

    /* Map same page twice */
    futex_a = mmap(NULL, page, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
    futex_b = mmap(NULL, page, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);

    if (futex_a == MAP_FAILED || futex_b == MAP_FAILED) {
        perror("mmap");
        exit(1);
    }

    printf("futex_a VA: %p\n", (void *)futex_a);
    printf("futex_b VA: %p\n", (void *)futex_b);

    atomic_store(futex_a, 0);

    pthread_t t1, t2;
    pthread_create(&t1, NULL, waiter, NULL);
    pthread_create(&t2, NULL, waker, NULL);

    pthread_join(t1, NULL);
    pthread_join(t2, NULL);

    return 0;
}
```
Output:

```
futex_a VA: 0x7f355b9eb000
futex_b VA: 0x7f355b9ea000
Waiter waiting on VA 0x7f355b9eb000
Waker writing via VA 0x7f355b9ea000
Waiter woken, value=1
```
Notice how the futexes are at different VAs but the fact that it did get woken up proves that they actually refer to the same waitqueue.  Another thing to consider is that different processes could very well have different waitqueues at the same VA, but we'd wrongly refer to the same waitqueue! 
*  Maybe change the key type here to `UA`, to be more explicit.

src/process/threading.rs
```diff
@@ -38,3 +63,94 @@ pub async fn sys_set_robust_list(head: TUA<RobustListHead>, len: usize) -> Resul
...
+            };
+
+            // First read – may fault and therefore is done outside the lock.
+            let current: u32 = copy_from_user(uaddr).await?;
+            if current != val {
+                return Err(KernelError::TryAgain);
+            }
+
+            // Second read – ensures the value is still what we expect *after*
+            // the wait-queue entry is visible to other threads.
```
hexagonal-sun: Not sure I completely get why we need two reads of the same value?

src/process/threading.rs
```diff
@@ -38,3 +63,94 @@ pub async fn sys_set_robust_list(head: TUA<RobustListHead>, len: usize) -> Resul
...
+            if copy_from_user(uaddr).await? != val {
+                return Err(KernelError::TryAgain);
+            }
+
+            wait_until(
+                waitq_arc.clone(),
+                |state| &mut state.wakers,
+                |state| {
+                    if state.wakeups > 0 {
```
hexagonal-sun, line 120: I quite like how you've used semaphore semantics here to store any pending wakeups.  This avoids having to worry about faulting while a lock is held.  I think in-time when we have `try_` variants of copy_from_user which aren't async we should refactor this to match the semantics decribed in the man-page:

```
Analogously to an atomic compare-and-exchange operation that potentially changes shared memory, blocking via a futex is an atomic compare-and-block operation.
```

usertest/src/main.rs
```diff
@@ -111,6 +111,60 @@ fn test_write() {
...
+            0,
+        );
+        if ret2 != -1 {
+            panic!("futex wait did not error out as expected");
+        }
+    }
+    println!(" OK");
+}
+
+fn test_rust_mutex() {
```
hexagonal-sun: Is this useful?  Given that the lock is uncontended, we futex isn't called at all!  I suppose to really test this, we need to get threading working (I'm looking into that at the moment).

arihant2math: > Is this useful? Given that the lock is uncontended, we futex isn't called at all! I suppose to really test this, we need to get threading working (I'm looking into that at the moment).

For some reason it actually did call futex, probably best to remove it though.

src/process/threading.rs
arihant2math: Ah, that was a bad copy and paste from the previous implementation.

PR #64, opened 2025-12-18
Initial ext4 support

arihant2math:
Redone version of #61. Uses a patched version of ext4-view which has been adapted to be async.

Advantages over fat32:
- soft/hard link support
- fast directory traversal
- larger files sizes

To test the changes:
```sh
# Move old image
mv moss.img fat_img.img
# Populate ext4_mount/ with the need stuff (just copy over from the fat mount) 
touch moss.img
mkfs.ext4 moss.img 256M
mkfs.ext4 -d ext4_mount moss.img
```

Then update `qemu-runner.sh` to change `--rootfs=fat32fs` to --rootfs=ext4fs`.

After this `cargo run -r` should work.

Fixes (partially): #4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,7 +92,7 @@ dependencies = [
 [[package]]
 name = "ext4-view"
 version = "0.9.3"
-source = "git+https://github.com/arihant2math/ext4-view-rs.git?branch=main#e63bccb01b2684df1dede05b7ac3330274d54d3a"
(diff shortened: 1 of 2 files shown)


arihant2math:
Finally, some progress. `ls` works!

arihant2math:
Seems like `ls` doesn't work, but `cat` seems to be working.
`readdir` seems to be the only barrier here.

PR #69, opened 2025-12-19
Clean exit of init

arihant2math:
exiting init (i.e. typing exit in bash) should lead to power off rather than a panic.

### DIFF
--- a/src/process/exit.rs
+++ b/src/process/exit.rs
@@ -1,19 +1,21 @@
-use crate::sched::current_task;
-use alloc::vec::Vec;
-use libkernel::error::Result;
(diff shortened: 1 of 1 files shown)


hexagonal-sun:
Are we sure we want to do this?  I'm pretty sure Linux panics when init exits... it should never happen.

IAmTheNerdNextDoor:
> Are we sure we want to do this? I'm pretty sure Linux panics when init exits... it should never happen.

yeah if init fails to run or exits, Linux does kernel panic.

arihant2math:
Ah I see, we just need a proper init process instead.

PR #70, opened 2025-12-20
Initial proc fs

arihant2math:
Provides an interface to userspace to list all running processes. Currently just shows their name/tid/state.

Adds functionality to keep track of the process name:
```
bash-5.3#  cat /proc/self/status
Name:   busybox
State:  R
Pid:    3
```

### DIFF
--- a/src/drivers/fs/proc.rs
+++ b/src/drivers/fs/proc.rs
@@ -1,10 +1,11 @@
 #![allow(clippy::module_name_repetitions)]
 
+use crate::process::thread_group::Tgid;
 use crate::sched::{SCHED_STATE, current_task};
+use crate::sync::OnceLock;
 use crate::{
     drivers::{Driver, FilesystemDriver},
     process::TASK_LIST,
-    sync::SpinLock,
 };
 use alloc::{boxed::Box, format, string::ToString, sync::Arc, vec::Vec};
 use async_trait::async_trait;
@@ -58,30 +59,29 @@ struct ProcDirStream {
 #[async_trait]
 impl DirStream for ProcDirStream {
     async fn next_entry(&mut self) -> Result<Option<Dirent>> {
-        if self.idx < self.entries.len() {
-            let ent = self.entries[self.idx].clone();
+        Ok(if let Some(entry) = self.entries.get(self.idx).cloned() {
             self.idx += 1;
-            Ok(Some(ent))
+            Some(entry)
         } else {
-            Ok(None)
-        }
+            None
+        })
     }
 }
 
 struct ProcRootInode {
     id: InodeId,
-    attr: SpinLock<FileAttr>,
+    attr: FileAttr,
 }
 
 impl ProcRootInode {
     fn new() -> Self {
         Self {
             id: InodeId::from_fsid_and_inodeid(PROCFS_ID, 0),
-            attr: SpinLock::new(FileAttr {
+            attr: FileAttr {
                 file_type: FileType::Directory,
                 mode: FilePermissions::from_bits_retain(0o555),
                 ..FileAttr::default()
-            }),
+            },
         }
     }
 }
@@ -94,36 +94,30 @@ impl Inode for ProcRootInode {
 
     async fn lookup(&self, name: &str) -> Result<Arc<dyn Inode>> {
         // Lookup a PID directory.
-        let pid: u32 = if name == "self" {
+        let pid = if name == "self" {
             let current_task = current_task();
-            current_task.descriptor().tid().0
+            current_task.descriptor().tgid()
         } else {
-            name.parse()
-                .map_err(|_| FsError::NotFound)
-                .map_err(Into::<KernelError>::into)?
+            Tgid(
+                name.parse()
+                    .map_err(|_| FsError::NotFound)
+                    .map_err(Into::<KernelError>::into)?,
+            )
         };
 
         // Validate that the process actually exists.
-        if !TASK_LIST
-            .lock_save_irq()
-            .keys()
-            .any(|d| d.tgid().value() == pid)
-        {
+        if !TASK_LIST.lock_save_irq().keys().any(|d| d.tgid() == pid) {
             return Err(FsError::NotFound.into());
         }
 
-        let fs = PROCFS_INSTANCE
-            .lock_save_irq()
-            .as_ref()
-            .expect("ProcFS singleton not initialised")
-            .clone();
+        let fs = procfs();
 
         let inode_id = fs.alloc_inode_id();
         Ok(Arc::new(ProcTaskInode::new(pid, inode_id)))
     }
 
     async fn getattr(&self) -> Result<FileAttr> {
-        Ok(self.attr.lock_save_irq().clone())
+        Ok(self.attr.clone())
     }
 
     async fn readdir(&self, start_offset: u64) -> Result<Box<dyn DirStream>> {
@@ -133,7 +127,8 @@ impl Inode for ProcRootInode {
         for (idx, desc) in task_list.keys().enumerate() {
             // Use offset index as dirent offset.
             let name = desc.tgid().value().to_string();
-            let inode_id = InodeId::from_fsid_and_inodeid(PROCFS_ID, (idx + 1) as u64);
+            let inode_id =
+                InodeId::from_fsid_and_inodeid(PROCFS_ID, ((desc.tgid().0 + 1) * 100) as u64);
             entries.push(Dirent::new(
                 name,
                 inode_id,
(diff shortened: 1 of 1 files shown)


scripts/qemu-runner.sh
```diff
@@ -8,4 +8,4 @@ bin="${elf%.elf}.bin"
 
 # Convert to binary format
 aarch64-none-elf-objcopy -O binary "$elf" "$bin"
-qemu-system-aarch64 -M virt,gic-version=3 -initrd moss.img -cpu cortex-a72 -m 2G -smp 4 -nographic -s -kernel "$bin" -append "--init=/bin/bash --rootfs=fat32fs --automount=/dev,devfs --automount=/tmp,tmpfs"
```
hexagonal-sun, line 11: Might be worth adding a line into `create-image.sh` too so that `/proc` is created in a fresh image.

arihant2math:
I think I fixed the directory layout issues earlier.
I've added `comm` and `state` for now, we need symlink support before adding cwd etc.
I've fixed the process reaping bug in f9f8e12fc255c28f4d7e537fc0f9921f6796f42b.

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+        }
+        entries.push(Dirent::new(
+            "self".to_string(),
+            InodeId::from_fsid_and_inodeid(PROCFS_ID, 0), // placeholder
+            FileType::Directory,
+            (entries.len() + 1) as u64,
+        ));
+
+        // honour start_offset
+        let entries = if (start_offset as usize) < entries.len() {
```
hexagonal-sun: I think you can just get away with:

```rust
        let entries = entries.into_iter().skip(start_offset as usize).collect();
```

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+}
+
+struct ProcDirStream {
+    entries: Vec<Dirent>,
+    idx: usize,
+}
+
+#[async_trait]
+impl DirStream for ProcDirStream {
+    async fn next_entry(&mut self) -> Result<Option<Dirent>> {
```
hexagonal-sun, line 61: Perhaps use `get()` here to remove boundary check logic:

```rust
    async fn next_entry(&mut self) -> Result<Option<Dirent>> {
        Ok(if let Some(entry) = self.entries.get(self.idx).cloned() {
            self.idx += 1;
            Some(entry)
        } else {
            None
        })
    }
```

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+        }
+    }
+
+    async fn getattr(&self) -> Result<FileAttr> {
+        Ok(self.attr.lock_save_irq().clone())
+    }
+
+    async fn readdir(&self, start_offset: u64) -> Result<Box<dyn DirStream>> {
+        let mut entries: Vec<Dirent> = Vec::new();
+        // Single entry: "status"
```
hexagonal-sun: Old comment?

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+        let inode_id = InodeId::from_fsid_and_inodeid(PROCFS_ID, 0); // placeholder
```
hexagonal-sun: So, I still was having issues when running:

```
bash-5.3# ./bin/busybox ls /proc/0
bash-5.3# ./bin/busybox ls -la /proc/0
total 0
bash-5.3#  
```

Yet, if i was to try and tab-complete in that directory, bash shows me the files as expected.  I figured out that the issue lies in reusing the root inode_id.  If I replace this with something like:

```rust
        entries.push(Dirent::new(
            "status".to_string(),
            InodeId::from_fsid_and_inodeid(PROCFS_ID, 3),
            FileType::File,
            1,
        ));
        entries.push(Dirent::new(
            "comm".to_string(),
            InodeId::from_fsid_and_inodeid(PROCFS_ID, 5),
            FileType::File,
            2,
        ));
        entries.push(Dirent::new(
            "state".to_string(),
            InodeId::from_fsid_and_inodeid(PROCFS_ID, 4),
            FileType::File,
            3,
        ));
```

Then ls works!  I suspect `ls` is getting confused, since we reuse the same inode id as the `/proc` directory and it would then interpret this has a hard-link back to a part of the path which it has already traversed.  Perhaps we've hit some sort of anti-cycle logic.

Ideally inode numbers should be stable - i.e. `stat /proc/1/comm` when ran multiple times should return the same inode number, but that requires storing inode numbers somewhere.  Once hack-y way you could do it would be to increment the inode allocator by say 100 for each top-level node in proc. That then gives each `/proc/<pid>` node 100 inode numbers to work with and you just allocate from that pool.

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+        entries.push(Dirent::new("comm".to_string(), inode_id, FileType::File, 2));
+        entries.push(Dirent::new(
+            "state".to_string(),
+            inode_id,
+            FileType::File,
+            3,
+        ));
+
+        // honour start_offset
+        let entries = if (start_offset as usize) < entries.len() {
```
hexagonal-sun: Ditto as to the other comment here.

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+}
+
+enum TaskFileType {
+    Status,
+    Comm,
+    State,
+}
+
+impl TaskFileType {
+    fn from_filename(name: &str) -> Option<Self> {
```
hexagonal-sun: Maybe it's more idiomatic to implement `TryInto` here?

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+            Ok(Some(ent))
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+struct ProcRootInode {
+    id: InodeId,
+    attr: SpinLock<FileAttr>,
```
hexagonal-sun: Any reason this is behind a spinlock?  AFAICS it's value is set on construction and then never changes.. I think we could drop the spinlock.

Likewise for the other structs.

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+            _ => None,
+        }
+    }
+}
+
+struct ProcTaskFileInode {
+    id: InodeId,
+    file_type: TaskFileType,
+    attr: SpinLock<FileAttr>,
+    pid: u32,
```
hexagonal-sun: Small point - I'd use a stronger type here: `Tgid`.

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+        Err(FsError::NotADirectory.into())
+    }
+
+    async fn read_at(&self, offset: u64, buf: &mut [u8]) -> Result<usize> {
+        let pid = self.pid;
+        let task_list = TASK_LIST.lock_save_irq();
+        let id = task_list
+            .iter()
+            .find(|(desc, _)| desc.tgid().value() == pid);
+        let task_details = if let Some((desc, _)) = id {
```
hexagonal-sun, line 292: I think I'd put a huge todo here.... the biggest challenge when writing procfs is obtaining details about a task which is not on this CPU's runqueue.  For now, we don't migrate tasks to other CPUs (although we are very close to doing that) so it's not a problem.   

Once we do implement task migration, my first thought on how to implement this would be with a specific IPI though the `CpuMessenger` module. That would  allow you to ask another CPU to get info about a task on that CPUs runqueue and return it.  Should be a lot of fun to implement!

src/drivers/fs/proc.rs
```diff
@@ -0,0 +1,381 @@
...
+        if end == 0 {
+            return Ok(0);
+        }
+        let slice = &bytes[offset as usize..offset as usize + end];
+        buf[..end].copy_from_slice(slice);
+        Ok(end)
+    }
+}
+
+static PROCFS_INSTANCE: SpinLock<Option<Arc<ProcFs>>> = SpinLock::new(None);
```
hexagonal-sun: Use `OnceLock`, that's the data structure we should use for singletons.

src/sched/mod.rs
```diff
@@ -70,7 +70,7 @@ pub fn insert_task(task: Arc<Task>) {
 
 pub struct SchedState {
     running_task: Option<Arc<Task>>,
-    run_queue: BTreeMap<TaskDescriptor, Arc<Task>>,
+    pub run_queue: BTreeMap<TaskDescriptor, Arc<Task>>,
```
hexagonal-sun, line 73: I'm not sure whether I like the run_queue and `SCHED_STATE` for that matter becoming `pub`.  Maybe write a public function within sched, something like `find_task`.  That could then contain the logic for getting it off the CPUs runqueue, if that's where it lives - or sending an IPI.

Having said that, I do wonder whether it's worth having just a huge `TASK_LIST`, which is a `BTreeMap<TaskDescriptor, Arc<Task>>`, for *all* tasks on the machine, not just this CPUs, that would make things easier I think.

arihant2math: I think linux uses per-cpu run-queues to reduce contention. Having a list of `SchedState`s would be helpful however.

Also it seems we're only scheduling new tasks on the previous task's CPU? Seems inefficient.

arihant2math: Basically hold a `Vec<SchedState>` and each CPU can have a pointer to it's specific sched state to eliminate that lookup.

hexagonal-sun: Ah, I didn't mean having a list of `SchedState`, I meant having a global `BTreeMap<TaskDescriptor, Arc<Task>>` of *all* tasks in the system.  That data structure sits alongside our existing per_cpu runqeues.

arihant2math: Ah ok, that might make sense.

PR #71, opened 2025-12-20
Implement sys_sysinfo

arihant2math:
Allows `uptime` to work, albeit with no load averages. To reduce conflict load calculations probably should be done after #70 is merged.


hexagonal-sun:
LGTM!

PR #73, opened 2025-12-21
Test usertests in CI

arihant2math:


arihant2math:
Interesting ... https://musl.cc blocks all microsoft IPs.

hexagonal-sun:
This would be awesome to have!  Plus we could even do ump and smp runs, ensuring the work in #91 doesn't regress.

arihant2math:
After thinking about it, I think it might be easier to set up a dockerfile. This would also potentially help anyone who has a windows host.

arihant2math:
Closing in favor of #109.

PR #75, opened 2025-12-22
Fix futex based thread joining

arihant2math:
Fixes the issue from https://github.com/hexagonal-sun/moss-kernel/issues/37#issuecomment-3679601382.

### DIFF
--- a/src/process/clone.rs
+++ b/src/process/clone.rs
@@ -1,3 +1,4 @@
+use super::{ctx::Context, thread_group::signal::SigSet};
 use crate::memory::uaccess::copy_to_user;
 use crate::{
     process::{TASK_LIST, Task, TaskState},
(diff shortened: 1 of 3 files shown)


src/process/clone.rs
```diff
@@ -44,8 +45,8 @@ bitflags! {
 pub async fn sys_clone(
     flags: u32,
     newsp: UA,
-    _arent_tidptr: UA,
```
hexagonal-sun, line 47: Could these be `TUA<u32>`?

src/process/exit.rs
```diff
@@ -101,6 +102,17 @@ pub fn sys_exit_group(exit_code: usize) -> Result<usize> {
 
 pub fn sys_exit(exit_code: usize) -> Result<usize> {
     let task = current_task();
+
+    // Honour CLONE_CHILD_CLEARTID: clear the user TID word and futex-wake any waiters.
+    if let Some(ptr) = task.child_tid_ptr.lock_save_irq().take() {
+        unsafe {
+            (ptr.value() as *mut u32).write_volatile(0);
```
hexagonal-sun: This should really be done through `copy_to_user`. Looks like this could be trivially converted to an async function since it's only called from the syscall interface.

PR #81, opened 2025-12-23
Support timeouts for futex

arihant2math:
The BITSET variant specifies a duration from the EPOCH, but the regular variant it just an ordinary duration.

### DIFF
--- a/src/clock/realtime.rs
+++ b/src/clock/realtime.rs
@@ -1,5 +1,5 @@
 use crate::{
-    drivers::timer::{Instant, now},
+    drivers::timer::{Instant, now, uptime},
     sync::SpinLock,
 };
 use core::time::Duration;
@@ -14,7 +14,7 @@ pub fn date() -> Duration {
         let duraton_since_ep_info = now - ep_info.1;
         ep_info.0 + duraton_since_ep_info
     } else {
-        Duration::new(0, 0)
+        uptime()
     }
 }
 


usertest/src/main.rs
```diff
@@ -174,44 +174,6 @@ fn test_write() {
     println!(" OK");
 }
 
-fn test_futex() {
```
hexagonal-sun, line 177: Any reason we've removed these tests?  I quite like having primitive tests that call the syscall.  I have another PR cooking which adds some edge-case tests too for things like a bogus uaddr, ensuring the fault handling mechanism works correctly, etc.

hexagonal-sun (approved):
Looks good!  Just one comment about removal of primitive tests.  Once `test_parking_lot_mutex_timeout` is uncommented, happy to merge.

usertest/src/main.rs
arihant2math: Sure, I've added it back in.

arihant2math:
Hmm, the parking lot test hangs, I don't think this is implemented quite right.

arihant2math:
Seems like `parking_lot` is retrying even when the syscall returns an error due to timeout?

hexagonal-sun:
My hunch is because of this:

```rust
fn test_clock_sleep() {
    const SLEEP_LEN: Duration = Duration::from_millis(100);

    print!("Testing clock and sleep syscalls ...");
    let now = Instant::now();
    sleep(SLEEP_LEN);
    dbg!(now.elapsed());
    assert!(now.elapsed() >= SLEEP_LEN);
    println!(" OK");
}
```

Even though this does actually sleep, but the duration returns:

```
[usertest/src/main.rs:20:5] now.elapsed() = 0ns

thread 'main' (4) panicked at usertest/src/main.rs:21:5:
assertion failed: now.elapsed() >= SLEEP_LEN
```

So parking_lot thinks it didn't sleep for any time at all,  therefore keeps trying until it has actually slept for 100ms.

arihant2math:
~~I think `CLOCK_REALTIME` is also just reporting uptime instead of time from epoch?~~
I'll see if I can fix this one, perhaps sleeping is broken, since the date function seems to work.

hexagonal-sun:
Okay, I have a fix:

```diff
diff --git a/src/clock/realtime.rs b/src/clock/realtime.rs
index 05f8a44..712210d 100644
--- a/src/clock/realtime.rs
+++ b/src/clock/realtime.rs
@@ -1,5 +1,5 @@
 use crate::{
-    drivers::timer::{Instant, now},
+    drivers::timer::{Instant, now, uptime},
     sync::SpinLock,
 };
 use core::time::Duration;
@@ -14,7 +14,7 @@ pub fn date() -> Duration {
         let duraton_since_ep_info = now - ep_info.1;
         ep_info.0 + duraton_since_ep_info
     } else {
-        Duration::new(0, 0)
+        uptime()
     }
 }
```

If the date epoch hasn't been set then we should fallback to uptime.  That fixes the test.

arihant2math:
I'll see if I can query an approximate time from the RTC in a different PR.

hexagonal-sun:
Yes, having proper RTC drivers would be good.

arihant2math:
Applied.

hexagonal-sun:
Perfect, if that fixes the tests on your machine, I'm happy to merge.

arihant2math:
👍 It does.

PR #83, opened 2025-12-23
Initial layout for networking syscalls

arihant2math:
Should be mergable after I fix the address copying.


arihant2math:
Closing due to me thinking of a different approach.

Issue #59, opened 2025-12-16
Ctrl-C doesn't work


arihant2math:
Terminating programs with Ctrl-C leads to no change/termination at all. `SIGINT` isn't being handled, therefore nothing happens.

PR #85, opened 2025-12-23
Make Ctrl-C et. al. dispatch signal

arihant2math:
I was trying to debug #59 and realized no signal is being dispatched at all. This doesn't completely fix that issue because it seems the signal isn't being handled properly.


hexagonal-sun (approved):
Looks good. Haven't managed to test it but happy to merge.

arihant2math:
It's not really testable beyond just seeing the Ctrl+C echo.

PR #91, opened 2025-12-25
Schedule onto multiple cores in round-robin fashion

arihant2math:
Exposes better APIs to collect all tasks/querying tasks across CPUs. Schedules in a round-robin ordering for now. Eventually we'd want work stealing from other cores and a quick way to get the current queue lengths for each core. Also this forces timer interrupts and extends them to non-main cores.

Currently fixes the bug where background tasks force a hang.

Includes improvements to the EEVDF implementation which fix many bugs.

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -1,5 +1,6 @@
 use crate::drivers::timer::Instant;
 use crate::process::threading::RobustListHead;
+use crate::sched::CpuId;
 use crate::{
     arch::{Arch, ArchImpl},
     fs::DummyInode,
@@ -14,7 +15,7 @@ use creds::Credentials;
 use ctx::{Context, UserCtx};
 use fd_table::FileDescriptorTable;
 use libkernel::memory::address::TUA;
-use libkernel::{CpuOps, VirtualMemory, fs::Inode};
+use libkernel::{VirtualMemory, fs::Inode};
 use libkernel::{
     fs::pathbuf::PathBuf,
     memory::{
@@ -193,7 +194,7 @@ pub struct Task {
     pub state: Arc<SpinLock<TaskState>>,
     pub robust_list: SpinLock<Option<TUA<RobustListHead>>>,
     pub child_tid_ptr: SpinLock<Option<TUA<u32>>>,
-    pub last_cpu: SpinLock<usize>,
+    pub last_cpu: SpinLock<CpuId>,
 }
 
 impl Task {
@@ -230,7 +231,7 @@ impl Task {
             last_run: SpinLock::new(None),
             robust_list: SpinLock::new(None),
             child_tid_ptr: SpinLock::new(None),
-            last_cpu: SpinLock::new(ArchImpl::id()),
+            last_cpu: SpinLock::new(CpuId::this()),
         }
     }
 
@@ -261,7 +262,7 @@ impl Task {
             last_run: SpinLock::new(None),
             robust_list: SpinLock::new(None),
             child_tid_ptr: SpinLock::new(None),
-            last_cpu: SpinLock::new(ArchImpl::id()),
+            last_cpu: SpinLock::new(CpuId::this()),
         }
     }
 
--- a/src/sched/mod.rs
+++ b/src/sched/mod.rs
@@ -323,7 +323,7 @@ impl SchedState {
         }
 
         *next_task.exec_start.lock_save_irq() = Some(now_inst);
-        *next_task.last_cpu.lock_save_irq() = CpuId::this().value();
+        *next_task.last_cpu.lock_save_irq() = CpuId::this();
 
         // Make sure the task possesses an eligible virtual deadline. If none is set
         // (or the previous one has elapsed), we hand out a brand-new one.
--- a/src/process/clone.rs
+++ b/src/process/clone.rs
@@ -1,6 +1,6 @@
 use super::{ctx::Context, thread_group::signal::SigSet};
-use crate::arch::ArchImpl;
 use crate::memory::uaccess::copy_to_user;
+use crate::sched::CpuId;
 use crate::{
     process::{TASK_LIST, Task, TaskState},
     sched::{self, current_task},
@@ -9,7 +9,6 @@ use crate::{
 use bitflags::bitflags;
 use libkernel::memory::address::TUA;
 use libkernel::{
-    CpuOps,
     error::{KernelError, Result},
     memory::address::UA,
 };
@@ -156,7 +155,7 @@ pub async fn sys_clone(
             } else {
                 None
             }),
-            last_cpu: SpinLock::new(ArchImpl::id()),
+            last_cpu: SpinLock::new(CpuId::this()),
         }
     };
 


src/sched/mod.rs
```diff
@@ -1,17 +1,90 @@
...
+        CpuId(ArchImpl::id())
+    }
+
+    pub fn value(&self) -> usize {
+        self.0
+    }
+}
+
+// TODO: arbitrary cap.
+pub static SCHED_STATES: [SyncUnsafeCell<Option<SchedState>>; 128] =
```
hexagonal-sun: I presume you’ve ditched the `per_cpu!` macro here since it precludes access to state held by other cores?

I’ve been thinking about this for a while and wondering what the best approach is. I can see two options:

* What you’ve done here: i.e., a static array which allows arbitrary access from any other core. The main downside I see is that everything must be locked, since it can be accessed by another core. I see that you’ve used `SyncUnsafeCell`; in that case, I think functions such as `get_sched_state_by_id` and `with_cpu_sched_state` are inherently unsafe, as they permit lock-free access to `SchedState` across CPUs.

* Accessing/modifying CPU state via the `CPUMessenger`. In this approach, we fire off an IPI to the target CPU, which performs the read/modify locally (lock-free) and returns the result.

I think the second option is a good compromise. Updating or accessing the state of another CPU isn’t done nearly as often as accessing the current CPU’s state. With the first option, we’d need locking for every scheduler state access or update (even something as frequent as retrieving the current task). Given that, my vote would be to try the` CPUMessenger` approach. I’m happy to give that a go if you’d like.

src/drivers/interrupts/arm_gic_v3.rs
```diff
@@ -411,6 +411,12 @@ impl InterruptController for ArmGicV3 {
                 trigger: TriggerMode::EdgeRising,
             });
         }
+
+        // Also enable the timer
```
hexagonal-sun, line 415: Shouldn't the timer already be enabled?  See [here](https://github.com/hexagonal-sun/moss-kernel/blob/8030cf0d8f7a87e23ea31e6549c6b9ec3624c740/src/drivers/timer/armv8_arch.rs#L72).

hexagonal-sun: Ah, I see this is to enable it for secondaries.  Ideally the driver should do this somehow, rather than relying on the gic driver to do it.

src/drivers/timer/mod.rs
```diff
@@ -132,17 +130,23 @@ impl InterruptHandler for SysTimer {
 
                 match event.what {
                     WakeupKind::Task(waker) => waker.wake(),
-                    WakeupKind::_Preempt => todo!(),
+                    WakeupKind::Preempt => crate::sched::sched_yield(),
```
hexagonal-sun: Could we do:

```rust
WakeupKind::Preempt => {
                wake_q.push(WakeupEvent {
                    self.driver.now() + Duration::from_millis(15),
                    what: WakeupKind::Preempt,
                });
                crate::sched::sched_yield();
}
```

i.e. the preempt wakeup always rechedules itself, then the below logic should be simplified and could probably revert to what was already there.

hexagonal-sun: Looking at the code further, perhaps not?  It looks as though the sched algorithm chooses a dynamic preemption point?

src/sched/mod.rs
hexagonal-sun: For an example of this, I've just pushed [this branch](https://github.com/hexagonal-sun/moss-kernel/compare/master...cpu-messenger-task-migration) which pushes tasks onto other CPUs via IPIs with the `CPUMessenger`.  When I run `usertest`, I'm getting deadlock at the moment:

```
bash-5.3# ./bin/usertest 
Running userspace tests ...
[   35.35031303] interrupts::cpu_messenger: Putting task on CPU 1
Testing sync syscall ... OK
```

I suspect this is the child process that's finished on CPU1, the waker seting the task in CPU0 to be runnable again, but it's stuck in the idle task.  I need to implement something such that a Waker, when signalling to another CPU causes the rechedule IPI message.

arihant2math: Yes. I've intentionally left out cross-cpu access beyond task insertion. To my knowledge get_sched_state_by_id et. al. should be safe since I'm never overwriting any item in the array, and the interior mutability of other threads is limited to `run_queue`, which is protected by a SpinLock.

A better approach could be to `Arc` the runqueue and just share that.

On the topic of using messengers, how fast are they?

src/drivers/timer/mod.rs
arihant2math: Yes, EEVDF doesn't fix a time quanta because of how it uses virtual time.

src/sched/mod.rs
arihant2math: I think the optimal cross-CPU task balancing method is work-stealing rather than work-giving, how would that work with messengers? It seems like you'd need to query for the number of tasks first, then request a steal.

src/drivers/interrupts/arm_gic_v3.rs
arihant2math: Yeah, that came up when I saw that schedule was never called on secondaries.

src/sched/mod.rs
hexagonal-sun: > Yes. I've intentionally left out cross-cpu access beyond task insertion. To my knowledge get_sched_state_by_id et. al. should be safe since I'm never overwriting any item in the array, and the interior mutability of other threads is limited to `run_queue`, which is protected by a SpinLock.

I disagree. I see a code path where the procfs module can call `find_task_by_descriptor`. That call can originate from any task on any CPU, which means it will very likely result in cross-CPU data access (i.e. accessing the `run_queue` of another CPU). That code iterates through all active slots and walks the `run_queue`s until it finds the task.

As far as I can see, this is highly unsafe and will lead to data races. SchedState itself is not protected by a mutex, nor are any of its fields.

If we can guarantee that the only core accessing a given piece of data is the CPU that owns it, then we can safely access it lock-free. That is precisely the guarantee provided by `per_cpu!`: there is no way to access data that does not belong to the current CPU. In this case, however, that guarantee does not hold. You need a different mechanism here—this is exactly the problem the `CPUMessenger` is trying to solve.

> A better approach could be to Arc the runqueue and just share that.

Possibly. I did consider wrapping all of this state in a large `Arc`, but that effectively means locking everything. Any on-CPU operation would then require taking a lock, which I think introduces too much overhead for the scheduler fast path.

> On the topic of using messengers, how fast are they?

IPIs are fast. You can hit additional latency if interrupts are disabled on the target CPU, but Linux uses them extensively in practice. For example, on my machine:

```
$ cat /proc/interrupts

[...]
 IWI:     103089     111648      51802      50290    3480843      50846      37640     101369   IRQ work interrupts
 RTR:          0          0          0          0          0          0          0          0   APIC ICR read retries
 RES:      53258      64925      39159      37986     143873      61857      38363      49915   Rescheduling interrupts
 CAL:    1434334    1127025     974039     838055     844695     993721     932218     876858   Function call interrupts
 TLB:     684503     742799     728184     629270     646185     829325     783279     722283   TLB shootdowns
 TRM:          0          0          0          0          0          0          0          0   Thermal event interrupts
 THR:          0          0          0          0          0          0          0          0   Threshold APIC interrupts
 DFR:          0          0          0          0          0          0          0          0   Deferred Error APIC interrupts
 MCE:          0          0          0          0          0          0          0          0   Machine check exceptions
 MCP:         66         65         65         65         65         65         65         65   Machine check polls
 ERR:          0
```

These are the various IPIs Linux uses, and as you can see, they are heavily exercised.

I've just pushed some additional code to the branch I mentioned earlier, this resolves the deadlock that I had and allows me to get through nearly all of `usertest` without any kind of preemption.  I'm just using IPIs to wake-up other CPUs when needed. 

> I think the optimal cross-CPU task balancing method is work-stealing rather than work-giving, how would that work with messengers? It seems like you'd need to query for the number of tasks first, then request a steal.

That’s a fair point, and I agree that work-stealing is generally the better default for schedulers.

With messengers, the model shifts slightly compared to shared-memory designs, but it still maps cleanly. The key idea is that ownership never changes implicitly: the stealing CPU never directly inspects or mutates another CPU’s run queue. Instead, it asks the owning CPU to do the inspection and transfer on its behalf. Concretely, the stealing CPU raises a message to a target CPU with the necessary parameters, and the target CPU decides what (if any) tasks it can push back.

hexagonal-sun:
I've just tried to run the usertest and it looks like you're getting the same problem as I am on my `cpu-messenger-task-migration` branch:

```
Testing rust threads ... OK
[    5.5169154] moss: Kernel panicked at src/arch/arm64/memory/fault.rs:121:17: Page fault defered error, SIGBUS on process
```

I Need to investigate what's going on there.

src/sched/mod.rs
arihant2math: Alright, I think we can worry about the work stealing implementation later, I'll switch what I have currently to use messengers.

src/sched/mod.rs
```diff
@@ -7,11 +8,51 @@ use crate::{
...
+    pub fn this() -> CpuId {
+        CpuId(ArchImpl::id())
+    }
+
+    pub fn value(&self) -> usize {
+        self.0
+    }
+}
+
+// TODO: arbitrary cap.
```
hexagonal-sun, line 31: Redundant comment now.

src/sched/mod.rs
```diff
@@ -45,7 +93,34 @@ fn schedule() {
...
     let previous_task = current_task();
     let mut sched_state = SCHED_STATE.borrow_mut();
+
+    // Bring the virtual clock up-to-date so that eligibility tests use the
+    // most recent value.
+    let now_inst = now().expect("System timer not initialised");
+    sched_state.advance_vclock(now_inst);
+
     let next_task = sched_state.find_next_runnable_task();
+    // if previous_task.tid != next_task.tid {
```
hexagonal-sun, line 103: I presume this is useful debug code? 

It'd be a shame to remove such code when it's useful for debugging.  Perhaps we should make use of `log`'s `release_max_level_info`  feature flags or similar, so that useful code isn't removed.

src/sched/mod.rs
```diff
@@ -59,27 +134,136 @@ pub fn spawn_kernel_work(fut: impl Future<Output = ()> + 'static + Send) {
...
-    SCHED_STATE
-        .borrow_mut()
-        .run_queue
-        .insert(task.descriptor(), task);
+    SCHED_STATE.borrow_mut().add_task(task);
+}
+
+pub fn insert_task_cross_cpu(task: Arc<Task>) {
+    let cpu = get_next_cpu();
+    message_cpu(cpu.value(), Message::PutTask(task)).expect("Failed to send task to CPU");
```
hexagonal-sun: Could slightly optimise here to use `insert_task` if we're inserting to the current CPU.

src/sched/mod.rs
```diff
@@ -191,6 +415,7 @@ pub fn sched_init() {
         .activate();
 
     *init_task.state.lock_save_irq() = TaskState::Running;
+    SCHED_STATE.borrow_mut().running_task = Some(idle_task.clone());
```
hexagonal-sun, line 429: Don't we want `init` to be the first task executed?

arihant2math:
FYI still debugging some weird crashes with usertest. Also there's a lot of intermittent hanging.

hexagonal-sun:
I found https://github.com/hexagonal-sun/moss-kernel/commit/50e5b0b2f56b9a549d231b4431f0b9cb33e2fda6 helped a lot with intermittent hanging. Note that you need the prior commit to the interrupt manager to help prevent a deadlock in the interrupt manager.

hexagonal-sun:
I’m seeing some lock-ups due to the interrupt manager not being able to handle interrupts reentrantly. That should be fixed if you cherry-pick https://github.com/hexagonal-sun/moss-kernel/commit/6ae237544dc99fda2a0cf35cd4b318089589a065.

I'm also seeing a lot of lock-ups in the allocator:

```
#0  spinning_top::spinlock::{impl#1}::lock (self=0xffff800001213170 <moss::arch::arm64::memory::HEAP_ALLOCATOR>) at src/spinlock.rs:57
#1  0xffff800000164f94 in lock_api::mutex::Mutex<spinning_top::spinlock::RawSpinlock, linked_list_allocator::Heap>::lock<spinning_top::spinlock::RawSpinlock, linked_list_allocator::Heap> (self=0xffff800001213170 <moss::arch::arm64::memory::HEAP_ALLOCATOR>)
    at /home/matthew/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lock_api-0.4.14/src/mutex.rs:213
#2  0xffff800000166144 in linked_list_allocator::{impl#4}::alloc (self=0xffff800001213170 <moss::arch::arm64::memory::HEAP_ALLOCATOR>, layout=...)
    at src/lib.rs:336
#3  0xffff800000151088 in moss::arch::arm64::memory::_::__rust_alloc (size=24, align=8) at src/arch/arm64/memory/mod.rs:21
#4  0xffff800000119dc8 in alloc::alloc::alloc (layout=...)
    at /home/matthew/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:95
#5  alloc::alloc::Global::alloc_impl (self=0x1, layout=..., zeroed=false)
    at /home/matthew/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:190
#6  0xffff800000119f04 in alloc::alloc::{impl#1}::allocate (layout=...)
    at /home/matthew/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:251
#7  alloc::alloc::exchange_malloc (size=24, align=8)
    at /home/matthew/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:352
#8  0xffff8000000e4790 in alloc::boxed::{impl#0}::new<moss::drivers::interrupts::arm_gic_v3::ArmGicV3InterruptContext> (x=...)
    at /home/matthew/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/boxed.rs:265
#9  moss::drivers::interrupts::arm_gic_v3::{impl#5}::read_active_interrupt (self=0xffffb00000000ce0) at src/drivers/interrupts/arm_gic_v3.rs:395
#10 0xffff800000066360 in moss::interrupts::InterruptManager::get_active_handler (self=0xffffb00000000d08) at src/interrupts/mod.rs:152
#11 0xffff800000066140 in moss::interrupts::InterruptManager::handle_interrupt (self=0xffffb00000000d08) at src/interrupts/mod.rs:161
#12 0xffff8000000e0a2c in moss::arch::arm64::exceptions::el1_irq_spx (state=0xffffb8000001a8d0) at src/arch/arm64/exceptions/mod.rs:125
#13 0xffffe000000002f4 in ?? ()
```

This makes total sense: the spinlock used in the allocator does not disable interrupts. Therefore, when an ISR is running, the heap spinlock may already be held. If an allocation occurs during the ISR, this will deadlock.

hexagonal-sun:
Okay, I've just pushed fixes for the heap allocator and the interrupt manager to this PR which should help.  I now get a successful boot to bash pretty much every time which means I think we've fixed interrupt locking issues.  I'm now hitting:

```
[    0.717370] drivers::interrupts::arm_gic_v3: GICv3: CPU interface for core 2 enabled.
[    0.718020] arch::arm64::boot::secondary: CPU 2 online.
[    0.719852] drivers::interrupts::arm_gic_v3: GICv3: Redistributor for core 3 (MPIDR=0x80000003) is awake.
[    0.720558] drivers::interrupts::arm_gic_v3: GICv3: CPU interface for core 3 enabled.
[    0.721818] arch::arm64::boot::secondary: CPU 3 online.
[    0.725946] interrupts::cpu_messenger: CPU 1 recieved ping from CPU 0
bash-5.3# [    1.1196409] moss: Kernel panicked at src/sched/mod.rs:335:13: assertion `left == right` failed
  left: Sleeping
 right: Runnable
```

~~Which looks like it's a scheduler related issue.~~  This was an incorrect assertion check which I added to the sched code.  I've pushed a commit to remove it.

arihant2math:
Seems like I can spawn 3 `cat /dev/zero &`s, then it starts to slow down. This is fine for now.

However usertest still doesn't run to completion. I think there might be a bug with respect to cross-cpu task completion.

arihant2math:
Also interrupts being scheduled on top of times when the borrow is acquired is turning out to be a problem. (i.e. `insert_task` is running and then the timer interrupt strikes. This seems to cause hangups a lot.

src/sched/mod.rs
```diff
@@ -24,14 +65,21 @@ pub mod waker;
...
-///    and performs the architecture-specific context switch.
+///   > `Runnable` for the old task, `Runnable` > `Running` for the new task)
+///   > and performs the architecture-specific context switch.
 ///
 /// # Returns
 ///
 /// Nothing, but the CPU context will be set to the next runnable task. See
 /// `userspace_return` for how this is invoked.
 fn schedule() {
+    if SCHED_STATE.try_borrow_mut().is_none() {
```
arihant2math, line 76: This is my "fix" for that, but it just returns. Sometimes this happens instantly, causing the idle task to be executed ad-nauseam.

hexagonal-sun:
Fixed a bunch of issues.  The thread exit logic was utterly wrong - we were sending a SIGCHLD to the parent for *every* thread exit; I'm surprised threading code worked prior to this PR!

Now we've introduced proper SMP, the fault handling code didn't handle page-fault collisions on two different CPUs.  It treated an `AlreadyMapped` condition as a hard error, see [709845c](https://github.com/hexagonal-sun/moss-kernel/pull/91/commits/709845c380268c70ce3de5407e0f2faae5faeaa8) for details.

I think we're getting closer to this being workable now.

arihant2math:
Still being very non-deterministic from my end. Usertest does pass sometimes. I suspect there might be some sort of race condition somewhere. From a black-box perspective it seems the algorithm is fine, so I suspect it's something else.

arihant2math:
Ah ok, I see that preempts are being scheduled but not executed (sometimes).

arihant2math:
~~Ok I see the problem, we should disable interrupts when modifying the wait queue because it might be locked during a timer interrupt.~~

Or something is just holding the wakeup_q and never dropping it.

Nevermind, I see the problem:
Essentially SYS_TIMER seems to be running on a random CPU. However `sched_yield` yields the current CPU's scheduler. So, if the yield was meant for a different CPU it'll never go through.

Indeed, modifying `WakeupEvent::preempt` to include the callers CPU and adding the following patch:
```rust
                    WakeupKind::Preempt(cpu_id) => {
                        if cpu_id == ArchImpl::id() {
                            crate::sched::sched_yield();
                        } else {
                            log::warn!(
                                "SysTimer: received preemption event for CPU {}, but running on CPU {}",
                                cpu_id,
                                ArchImpl::id()
                            );
                        }
                    }
                }
```

Shows:
```
[    2.2638996] drivers::timer: SysTimer: received preemption event for CPU 1, but running on CPU 0
```
right before a hang.

Ideally we'd have some sort of per-cpu handling (either the wakeup-queue being per-cpu or the timer interrupts being handled per-cpu). But for now I'll just send an IPI.

Still running into the timer interrupt not firing (I think it's still being deadlocked somehow).

Ah, it's a cyclic lock acquire (sched_yield -> schedule -> schedule_preempt).

arihant2math:
Ok, fixed the hanging bug with 8a09fe0bca0b2803cd979d859b203ba725381824. Essentially I move to interrupt the idle task every 4 ms.

However I'm seeing a speed regression of ~10x when running usertests. I'd like to probably resolve that before merging.

hexagonal-sun:
Do you think we need a commit like 50e5b0b to force other CPUs out of the idle task if a waker changes the state of the task?  I realise that the preempt code would eventually do this with a tick interrupt but I perhaps it may reduce latency?

arihant2math:
> Do you think we need a commit like [50e5b0b](https://github.com/hexagonal-sun/moss-kernel/commit/50e5b0b2f56b9a549d231b4431f0b9cb33e2fda6) to force other CPUs out of the idle task if a waker changes the state of the task? I realise that the preempt code would eventually do this with a tick interrupt but I perhaps it may reduce latency?

This would help. I think we need a good way of ensuring that a child can quickly wake up it's parent.

arihant2math:
Removing cross-cpu scheduling by forcing everything to turn into single cpu leads to usertest taking 550ms. `master` only takes 356ms.

I'm looking at some optimizations like keeping threads on the same CPU to unless the local queue is too long.

I'm also thinking that we can improve on the way IPIs are being used by 1) batching new task sends and 2) using a lock-free `ArrayQueue` to push new tasks into a remote CPU.

I think the IPI stuff might be far-fetched for now, but I would be interested in somehow beating down that single-cpu number.

hexagonal-sun:
I'll have a look today.  I think the first thing would be to make the timer code be per-cpu.  We need a per-cpu `wakeup_q`.  That should allow preemption events to be scheduled for *this* CPU.  This will also allow `sleep()`s to be inserted on the work_q for this CPU.  I suspect this will help a lot.

IPIs should really be used for slow-path code.  Ideally we should minimise their use for the scheduling fast-path as much as possible.

hexagonal-sun:
Okay, this is starting to look more reasonable for me now.  It seems much more deterministic, I've not noticed any hangs or anything of that sort.  Also `usertest` times between smp and ump seem comparable now:

smp (4 cpus):
```
All tests passed in 786 ms
All tests passed in 717 ms
All tests passed in 716 ms
```

ump:

```
All tests passed in 908 ms
All tests passed in 794 ms
All tests passed in 882 ms
```

@arihant2math Do you think it's worth us both having a code review pass at this point?  If we're happy, we could look at merging.  There are some other bits of work that I'd like to spawn off from this, but they could be addressed in separate PRs (this one is already getting fairly large).

arihant2math:
> @arihant2math Do you think it's worth us both having a code review pass at this point? If we're happy, we could look at merging. There are some other bits of work that I'd like to spawn off from this, but they could be addressed in separate PRs (this one is already getting fairly large).

Yeah, I think it would be good to review now, seeing that everything seems to be deterministic and relatively performant. If there aren't any glaring issues we should probably look at merging it.

src/sched/mod.rs
arihant2math: I hope with a proper procfs we can remove this and rely on top instead.

src/sched/mod.rs
```diff
@@ -7,11 +8,50 @@ use crate::{
...
+pub const VT_ONE: u128 = 1u128 << VT_FIXED_SHIFT;
+/// Tolerance used when comparing virtual-time values (see EEVDF, Fixed-Point Arithmetic).
+/// Two virtual-time instants whose integer parts differ by no more than this constant are considered equal.
+pub const VCLOCK_EPSILON: u128 = VT_ONE;
+
+pub fn find_task_by_descriptor(descriptor: &TaskDescriptor) -> Option<Arc<Task>> {
+    if let Some(task) = SCHED_STATE.borrow().run_queue.get(descriptor) {
+        return Some(task.clone());
+    }
+    // TODO: Ping other CPUs to find the task.
```
arihant2math, line 52: Should probably be handled by an IPI in a future PR

src/interrupts/cpu_messenger.rs
```diff
@@ -47,8 +47,7 @@ impl InterruptHandler for CpuMessenger {
             .try_pop();
 
         match message {
-            // Some(Message::Reschedule) => return, // We reschedule when returning from an IRQ.
-            // Some(Message::PutTask(task)) => sched::insert_task(task),
+            Some(Message::PutTask(task)) => sched::insert_task(task),
```
arihant2math, line 50: Might be worth calling a reschedule here in the future to reduce latency (if it does).

src/drivers/fs/proc.rs
```diff
@@ -305,11 +304,11 @@ State:\t{state}
 Tgid:\t{tgid}
 FDSize:\t{fd_size}
 Pid:\t{pid}
-Threads:\t{threads}\n",
+Tasks:\t{tasks}\n",
```
arihant2math: Is this right? I'm pretty sure it's threads.

src/drivers/timer/mod.rs
```diff
@@ -163,19 +166,19 @@ impl SysTimer {
...
+
+                wakeup_q.push(WakeupEvent {
                     when,
                     what: WakeupKind::Task(cx.waker().clone()),
                 });
 
                 // After pushing, we must update the hardware timer in case our
                 // new event is the earliest one.
-                if let Some(next_event) = wake_q.peek() {
+                if let Some(next_event) = wakeup_q.peek() {
```
arihant2math, line 181: This might stop being accurate now that schedule() is called after this.

src/drivers/timer/mod.rs
```diff
@@ -213,4 +247,38 @@ pub async fn sleep(duration: Duration) {
...
+
+/// Arms a pre-emption timer for the running task on this CPU.
+/// Called by the scheduler every time it issues a new eligible virtual deadline.
+pub fn schedule_preempt(when: Instant) {
+    if let Some(timer) = SYS_TIMER.get() {
+        timer.schedule_preempt(when);
+    }
+}
+
+pub fn schedule_force_preempt() {
```
arihant2math, line 266: We should find a way to do without this later.

src/process/threading/futex/mod.rs
```diff
@@ -80,18 +80,19 @@ pub async fn sys_futex(
...
-        let timeout = TimeSpec::copy_from_user(timeout).await?;
-        if matches!(cmd, FUTEX_WAIT_BITSET) {
-            Some(Duration::from(timeout) - date())
-        } else {
-            Some(Duration::from(timeout))
-        }
-    };
     match cmd {
         FUTEX_WAIT | FUTEX_WAIT_BITSET => {
+            let timeout = if timeout.is_null() {
```
arihant2math, line 85: Might be worth undoing this and moving this to the next futex PR.

src/drivers/timer/mod.rs
arihant2math: Although it doesn't matter because we'll probably remove this in the near-future anyhow.

src/drivers/fs/proc.rs
arihant2math: Also, does the refactor from threads -> tasks make sense? imo threads is more intuitive. It's fine either way.

src/interrupts/cpu_messenger.rs
arihant2math: Alternatively it might make sense to call a reschedule after any insert_task.

src/sched/mod.rs
hexagonal-sun: Agreed, althogh I've got a patch in the works which makes `TASK_LIST` a global list of all tasks, not just task states which should help with this.

src/interrupts/cpu_messenger.rs
hexagonal-sun: A reschedule should already occur.  Whenever we return from an exception (IRQ or syscall), we always do a reschedule.  Note that calling `schedule()` itself can be dangerous as it does do a context switch, so the caller *must* ensure that we're not touching any task-related state intended for the previous task after calling `schedule()`.

src/drivers/fs/proc.rs
hexagonal-sun: You are right, it should be `Threads`.  I'll change that.

src/process/threading/futex/mod.rs
hexagonal-sun: Agreed, there are a number of fixes related to other pieces of the kernel and don't have anything to do with scheduling.  The commits that are relatively isolated, I will pull out an raise separate PRs for them.  That should help reduce the size of this PR and make it more manageable.

src/drivers/timer/mod.rs
hexagonal-sun: Agreed, this feels like a bit of a hack.

hexagonal-sun:
I've split out independent fixes into sepereate PRs which has reduced the size of this PR.  Ideally they need merging before this does. They are #108 #107 #106 #105 #104.

src/sched/mod.rs
arihant2math: That works as well.

arihant2math:
I think the sub-PRs all look reasonable; I've merged them all. I also cleaned up my commit logs a little bit.
I think that merging this after your review, given that there are no issues, looks reasonable

src/process/mod.rs
```diff
@@ -176,14 +181,19 @@ pub struct Task {
...
+    /// Virtual deadline (v_di) used by the EEVDF scheduler.
+    pub v_deadline: SpinLock<u128>,
     pub exec_start: SpinLock<Option<Instant>>,
     pub deadline: SpinLock<Option<Instant>>,
     pub priority: i8,
     pub last_run: SpinLock<Option<Instant>>,
     pub state: Arc<SpinLock<TaskState>>,
     pub robust_list: SpinLock<Option<TUA<RobustListHead>>>,
     pub child_tid_ptr: SpinLock<Option<TUA<u32>>>,
+    pub last_cpu: SpinLock<usize>,
```
hexagonal-sun: Maybe we could use your new `CpuId` type here?

hexagonal-sun (approved):
One minor comment, but happy to merge.  I don't fully understand the EEVDF scheduling aspects but it seems to work fine.  I should set some time aside and study the paper really.

PR #109, opened 2025-12-29
Add docker env and test usertests on CI

arihant2math:
This is meant for GitHub CI but can be used for personal development as well.

### DIFF
--- a/scripts/build-deps.sh
+++ b/scripts/build-deps.sh
@@ -14,10 +14,23 @@ else
     MUSL_CC="aarch64-linux-musl-cross"
 fi
 
-if [ ! -f "${MUSL_CC}.tgz" ]
-then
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:latest
+
+# Install dependencies
+RUN apt update
+RUN apt install -y build-essential curl git wget
+RUN apt install -y qemu-system-aarch64 dosfstools mtools
+
+# Install Rust
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
arihant2math, line 9: Maybe we should only install nightly in a future PR. Also installing the required toolchains manually would help with caching.

PR #110, opened 2025-12-29
Update non-breaking dependencies

arihant2math:
Bumps all dependencies to latest versions except for `fdt-parser`.


hexagonal-sun:
Looks good.  Would be nice to wait for #109 to be merged so we can run this through the extended CI tests.

arihant2math:
I've rebased and it passes. It might be useful to make that a required check.

PR #114, opened 2025-12-29
Turn smp into a feature

arihant2math:
In order to test both single core and multi core scheduling I've turned SMP into a cargo feature which can be toggled. This allows for proper testing in CI of both UP and SMP.


arihant2math:
Cool, the feature config setup seems to work now.

Issue #94, opened 2025-12-26
Consider a `OwnedTask` struct


hexagonal-sun:
Implement some kind of `OwnedTask` struct.   I've been thinking about this one for a while.  We currently throw around a `Arc<Task>` as our main entry-point to tasks.  I see implementing `OwnedTask` has two benefits:

* It allows us to to keep track of ownership semantics in the type-system.
* We could provide lock-free access to fields that don't need to be shared (`ctx`, `pending_signals`, `creds`, etc).   Then, we could have a huge task lookup-table `BTreeMap<TaskDescriptor, Arc<Task>>` which would allow cross CPU data access to shared fields.

PR #116, opened 2025-12-30
process: split `Task` into `Task` (shared) and `OwnedTask` (local)

hexagonal-sun:
```
This commit refactors the core process representation to decouple
"Identity/Resources" from "Execution/Scheduling". Previously, a
monolithic `Task` struct wrapped in `Arc<SpinLock<>>` caused lock
contention during hot scheduling paths and conflated shared state with
CPU-local state.

The `Task` struct has been split into:

1. `Task` (Shared): Holds process-wide resources (VM, FileTable,
Credentials). Managed via `Arc` and internal fine-grained locking.

2. `OwnedTask` (Private): Holds execution state (Context, v_runtime,
signal mask). Strictly owned by a specific CPU (via the Scheduler) and
accessed lock-free.

Key changes:

* Scheduler:
  chedState` now owns tasks via `Box<OwnedTask>`.
  - Transitions between `run_queue` and `running_task` involve strictly
    moving ownership of the Box, ensuring pointer stability.
  - The EEVDF comparison logic now explicitly handles comparisons
    between the queued candidates and the currently running task (which is
    not in the queue).

* Current Task Access:
  - `current()` now returns a `CurrentTaskGuard` which:
    1. Disables preemption (preventing context switches while holding
       the reference).
    2. Performs a runtime borrow check (panic on double-mutable borrow).
    3. Dereferences a cached Per-CPU raw pointer for O(1) access.
```

Also includes:

    sched: implement fast-path return
    
    Previously, the scheduler unconditionally performed a full runqueue
    search:(`find_next_runnable_desc`) on every invocation, including every
    timer tick and syscall return. This resulted in unnecessary overhead.
    
    This change introduces a "lazy preemption" model:
    
    1. Fast-Path Optimization: `do_schedule` now checks if the current task
       is valid, is not the Idle task, and still has virtual budget remaining.
       If these conditions are met and `force_resched` is not set, the
       scheduler returns immediately without locking the runqueue.
    
    2. Preemption & Idle Handling:
       - `insert_into_runq` now sets `force_resched` if the new task has an
         earlier deadline than the current task, or if the current task is
         Idle.
       - The Idle task is explicitly excluded from the fast-path to ensure
         immediate context switching when new work arrives.

This is now at a point where it's ready for review.  @some100 this also touches some of the FS changes you've been making recently, might be worth having a quick look.  Thanks!

Fixes: #94

### DIFF
--- a/src/arch/arm64/boot/mod.rs
+++ b/src/arch/arm64/boot/mod.rs
@@ -10,10 +10,7 @@ use crate::{
         fdt_prober::{probe_for_fdt_devices, set_fdt_va},
         init::run_initcalls,
     },
-    interrupts::{
-        cpu_messenger::{Message, cpu_messenger_init, message_cpu},
-        get_interrupt_root,
-    },
+    interrupts::{cpu_messenger::cpu_messenger_init, get_interrupt_root},
     kmain,
     memory::{INITAL_ALLOCATOR, PAGE_ALLOC},
     sched::{sched_init_secondary, uspc_ret::dispatch_userspace_task},
@@ -132,8 +129,6 @@ fn arch_init_stage2(frame: *mut ExceptionState) -> *mut ExceptionState {
     boot_secondaries();
 
     // Prove that we can send IPIs through the messenger.
-    let _ = message_cpu(1, Message::Ping(ArchImpl::id() as _));
-
     frame
 }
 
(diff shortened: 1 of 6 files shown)


src/arch/arm64/boot/memory.rs
```diff
@@ -21,7 +21,7 @@ use libkernel::{
 };
 use log::info;
 
-const KERNEL_STACK_SZ: usize = 64 * 1024; // 32 KiB
+const KERNEL_STACK_SZ: usize = 256 * 1024; // 32 KiB
```
arihant2math, line 24: nit: Outdated comment

PR #120, opened 2026-01-02
process: task_group: implement pg priority

hexagonal-sun:
Currently, each task implements it's own priority value. In Linux, each
thread group (process) has a default process which all tasks in that
group inherit. Tasks can, however, override the default process
priority.

Implement that logic here which also fixes the current compilation error
on master.


arihant2math (approved):

arihant2math:
We should probably implement priority inheritance later.

PR #122, opened 2026-01-03
Use an ext4 rootfs with symlinks to busybox applets

### LINKED ISSUE #10: What shell to focus on
After playing around with busybox and the whole build for a while, it's becoming clear to me that everything is a bit more complex than expected. I assumed that busybox shell had access to all other busbox commands, but apparently that is not the case, which makes the busybox shell way less useful. In addition, busybox needs at least the syscall 0x63 to work.

So as long as the kernel cant use a file system that supports symlinks, we can either make a copy for each tool that busybox supports or accept that we don't have those tools for now. We probably could do some sort of hack to pretend those tools exist, while somehow mapping them to busybox, but I don't think that's the best idea.

If we decide to just accept that those tools don't exist for now, then we might as well move back to bash, which was already shown to work, but use the proper repo instead of the mirror.

Maybe there is also another option I'm not seeing.

I have a branch that can produce an image with just the pre-built busybox, which also moves the scripts into their own directory like described in #7, so if the decision falls on busybox, I can ready the PR for that. If the decision falls on bash, I can quickly re-work the branch and add the bash build again. I'm just not quite sure what the best way forward is.

some100:
Since there's now a working ro ext4 driver, there can now be symlinks for busybox applets. This means that it would no longer be a requirement to prefix almost every command with busybox.

this also maybe resolves #10


arihant2math (changes requested):
Can the `mac-experimental` build scripts also be updated?

some100:
Added support for mac-experimental, but I can't guarantee it to work since I haven't tested it on a mac. In theory it should probably work since e2fsprogs is also available on macOS.

arihant2math (approved):

PR #127, opened 2026-01-05
Socket implementation

arihant2math:
Implements socket syscalls (i.e. `socket`, `accept`, `connect`, `bind`, etc.) and unix sockets. Also stubs out part of a TCP socket implementation.

### DIFF
--- a/src/socket/mod.rs
+++ b/src/socket/mod.rs
@@ -4,7 +4,7 @@ mod tcp;
 mod unix;
 
 use crate::drivers::timer::now;
-use crate::memory::uaccess::copy_from_user;
+use crate::memory::uaccess::{copy_from_user, copy_from_user_slice};
 use crate::sync::OnceLock;
 use crate::sync::SpinLock;
 use alloc::vec;
@@ -138,10 +138,14 @@ pub async fn parse_sockaddr(uaddr: UA, len: usize) -> Result<SockAddr, KernelErr
             Ok(SockAddr::In(sain))
         }
         AF_UNIX => {
-            if len < size_of::<SockAddrUn>() {
+            let path_len = len - size_of::<u16>() * 2;
+            if path_len > 108 {
                 return Err(KernelError::InvalidValue);
             }
-            let saun: SockAddrUn = try_copy_from_user(uaddr.cast())?;
+            let mut path = [0u8; 108];
+            copy_from_user_slice(uaddr.add_bytes(size_of::<u16>()), &mut path[..path_len]).await?;
+            // print str of path
+            let saun: SockAddrUn = SockAddrUn { family, path };
             Ok(SockAddr::Un(saun))
         }
         _ => Err(KernelError::AddressFamilyNotSupported),
--- a/src/socket/unix.rs
+++ b/src/socket/unix.rs
@@ -12,7 +12,7 @@ use async_trait::async_trait;
 use core::future::poll_fn;
 use core::task::Poll;
 use core::task::Waker;
-use libkernel::error::{KernelError, Result};
+use libkernel::error::{FsError, KernelError, Result};
 use libkernel::memory::address::UA;
 
 /// Registry mapping Unix socket path bytes to endpoint inbox and listening state
@@ -133,7 +133,7 @@ impl SocketOps for UnixSocket {
                 };
                 let mut reg = endpoints().lock_save_irq();
                 let Some(ep) = reg.get_mut(&path) else {
-                    return Err(KernelError::InvalidValue);
+                    return Err(KernelError::Fs(FsError::NotFound));
                 };
                 if ep.listening {
                     if ep.pending.len() >= ep.backlog_max {


src/fs/fops.rs
```diff
@@ -137,4 +137,12 @@ pub trait FileOps: Send + Sync {
     ) -> Result<usize> {
         Err(KernelError::InvalidValue)
     }
+
+    fn is_socket(&self) -> bool {
```
hexagonal-sun: Do we need this function?  Could this not be `as_socket().is_some()`?

src/socket/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+static SOCKETS: OnceLock<SpinLock<SocketSet>> = OnceLock::new();
+
+fn sockets() -> &'static SpinLock<SocketSet<'static>> {
+    SOCKETS.get_or_init(|| SpinLock::new(SocketSet::new(vec![])))
+}
+
+// static INTERFACE: OnceLock<SpinLock<EthernetInterface<OurDevice>>> = OnceLock::new();
+// static DHCP_CLIENT: OnceLock<SpinLock<Dhcpv4Client>> = OnceLock::new();
+
+static DHCP_ENABLED: OnceLock<bool> = OnceLock::new();
```
hexagonal-sun: Should the kernel have anything to do with DHCP?  I think that should be left to userspace.  At least for ipv4.

src/socket/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+unsafe impl crate::memory::uaccess::UserCopyable for SockAddrUn {}
+
+impl TryFrom<SockAddr> for IpEndpoint {
+    type Error = KernelError;
+    fn try_from(sockaddr: SockAddr) -> Result<IpEndpoint, KernelError> {
+        match sockaddr {
+            SockAddr::In(SockAddrIn { port, addr, .. }) => Ok(IpEndpoint {
+                port: u16::from_be_bytes(port),
+                addr: if u32::from_be_bytes(addr) == 0 {
+                    IpAddress::Ipv4(Ipv4Addr::UNSPECIFIED)
```
hexagonal-sun: I think the if is redundant? If we convert '0' from bits, it should be unspecified anyway.

src/socket/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+pub fn parse_sockaddr(uaddr: UA, len: usize) -> Result<SockAddr, KernelError> {
+    use crate::memory::uaccess::try_copy_from_user;
+    use libkernel::memory::address::TUA;
+
+    // Need at least a family field
+    if len < size_of::<u16>() {
+        return Err(KernelError::InvalidValue);
+    }
+
+    let family: u16 = try_copy_from_user(TUA::from_value(uaddr.value()))?;
```
hexagonal-sun: Any reason we're using the try_copy variations here? I think this function is called exclusively from async fns, so it could be async allowing you to use the proper async copy fns.

src/socket/tcp.rs
```diff
@@ -0,0 +1,123 @@
...
+pub struct TcpSocket {
+    handle: SocketHandle,
+    local_endpoint: SpinLock<Option<IpEndpoint>>,
+    backlogs: SpinLock<Vec<Arc<TcpSocket>>>,
+    num_backlogs: AtomicUsize,
+}
+
+impl TcpSocket {
+    pub fn new() -> Self {
+        let rx_buffer = SocketBuffer::new(vec![0; 4096]);
```
hexagonal-sun, line 40: Rather than using the heap, maybe dedicate a `ClaimedPage` here.

src/socket/tcp.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn refill_backlog_sockets(
+        &self,
+        backlogs: &mut SpinLockIrqGuard<Vec<Arc<TcpSocket>>, ArchImpl>,
+    ) -> Result<(), KernelError> {
+        let local_endpoint = match *self.local_endpoint.lock_save_irq() {
+            Some(local_endpoint) => local_endpoint,
+            None => return Err(KernelError::InvalidValue),
+        };
+
+        for _ in 0..(self.num_backlogs.load(Ordering::SeqCst) - backlogs.len()) {
```
hexagonal-sun: Could probably get away with `Ordering::Relaxed` here.

src/socket/tcp.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub fn new() -> Self {
+        let rx_buffer = SocketBuffer::new(vec![0; 4096]);
+        let tx_buffer = SocketBuffer::new(vec![0; 4096]);
+        let inner = smoltcp::socket::tcp::Socket::new(rx_buffer, tx_buffer);
+        let handle = sockets().lock_save_irq().add(inner);
+        TcpSocket {
+            handle,
+            local_endpoint: SpinLock::new(None),
+            backlogs: SpinLock::new(Vec::new()),
+            num_backlogs: AtomicUsize::new(0),
```
hexagonal-sun, line 48: I'm not sure I quite get why we track this separately? Couldn't we call backlogs.len()?

src/socket/unix.rs
```diff
@@ -0,0 +1,284 @@
...
+    inbox: Arc<KPipe>,
+    listening: bool,
+    backlog_max: usize,
+    pending: Vec<UnixSocket>,
+    /// Wakers for tasks waiting in accept
+    waiters: Vec<Waker>,
+}
+
+/// Registry mapping Unix socket path bytes to endpoint inbox
+static UNIX_ENDPOINTS: OnceLock<SpinLock<BTreeMap<Vec<u8>, Endpoint>>> = OnceLock::new();
```
hexagonal-sun: Could we not use PathBuf as a key here? Or even better, an InodeId? Or even the sockaddrun?

src/socket/unix.rs
```diff
@@ -0,0 +1,284 @@
...
+        if end == 0 {
+            None
+        } else {
+            Some(saun.path[..end].to_vec())
+        }
+    }
+}
+
+#[async_trait]
+impl SocketOps for UnixSocket {
```
hexagonal-sun, line 99: I almost wonder whether we make a bit of an architectural change here and consider splitting the trait into two types.. a listening socket and a connected socket? Feels like this might make things easier.

hexagonal-sun: Then the registry would *only* contain listening sockets.

src/socket/unix.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    async fn read(&mut self, _ctx: &mut FileCtx, buf: UA, count: usize) -> Result<usize> {
+        if count == 0 {
+            return Ok(0);
+        }
+        if *self.rd_shutdown.lock_save_irq() {
+            return Ok(0);
+        }
+        self.inbox.copy_to_user(buf, count).await
```
hexagonal-sun, line 237: Once thing to note, `SeqPacket` and `Datagram` sockets will not preserve message boundaries with the kbuf, it's just a stream of bytes; a perfect fit for `Stream` sockets.

We need to think about how we can preserve message boundaries.

src/socket/tcp.rs
arihant2math: I was trying to reduce calls into the `SpinLock` for performance.

src/socket/tcp.rs
arihant2math: SocketBuffer needs something that can be converted into a `ManagedSlice`, so I have to give it an owned vec. Trying to give it a borrowed slice runs into issues because the slice would be given a lifetime of `'static`

PR #131, opened 2026-01-06
Mark freed pages as freed in munmap

arihant2math:
Previously munmap wouldn't marked freed pages as free, leading to a memory leak.


hexagonal-sun:
Although this does remind me, I do need to do full page reclimation when tearing down an address space.

PR #132, opened 2026-01-06
Set default path in kernel

arihant2math:
Not sure if this is the right solution, but it does the trick.


arihant2math:
Definitely not correct, closing.

PR #141, opened 2026-01-08
syscalls: wait4: implement WNOHANG

hexagonal-sun:
Implement WNOHANG logic within the sys_wait4 system call. If this is
flag is set, remove the process from the ChildNotifiers set without
waiting.

### DIFF
--- a/src/process/thread_group/wait.rs
+++ b/src/process/thread_group/wait.rs
@@ -97,6 +97,34 @@ impl ChildNotifiers {
     }
 }
 
+fn do_wait(
+    state: &mut BTreeMap<Tgid, ChildState>,
+    pid: PidT,
+    flags: WaitFlags,
+) -> Option<(Tgid, ChildState)> {
+    let key = if pid == -1 {
+        state.iter().find_map(|(k, v)| {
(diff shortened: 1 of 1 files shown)


src/process/thread_group/wait.rs
```diff
@@ -97,6 +97,34 @@ impl ChildNotifiers {
...
+            .and_then(|(k, v)| {
+                if v.matches_wait_flags(flags) {
+                    Some(*k)
+                } else {
+                    None
+                }
+            })
+    }?;
+
+    Some(state.remove_entry(&key).unwrap())
```
arihant2math, line 125: If there are multiple waiters, wouldn't this panic?

hexagonal-sun: No. The `BTreeMap` is protected inside a `SpinLock` within the `CondVar`, so this whole function is basically a critical section.  This guarantees that the the `unwrap()` is safe since we found the key above.

hexagonal-sun:
Is the CI failure spurious?

hexagonal-sun:
> Is the CI failure spurious?

Just re-ran and it now passes, so I think this was the case.

PR #144, opened 2026-01-10
Implement rwlock

arihant2math:


libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,223 @@
...
+enum RwlockLockStateInner {
+    Read(usize),
+    Write,
+}
+
+struct RwlockState {
+    is_locked: Option<RwlockLockStateInner>,
+    read_waiters: VecDeque<Waker>,
+    write_waiters: VecDeque<Waker>,
+    last_woken_was_writer: AtomicBool,
```
hexagonal-sun: No point this being atomic if it's protected by the `SpinLock`.

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,223 @@
...
+
+use super::spinlock::SpinLockIrq;
+
+enum RwlockLockStateInner {
+    Read(usize),
+    Write,
+}
+
+struct RwlockState {
+    is_locked: Option<RwlockLockStateInner>,
```
hexagonal-sun: I wonder whether it's trying to write this with an i32:

- `-1`: write locked
- `0`: unlocked
- `n > 0`, read-locked with `n` readers.

If we use that approach then we could look at using an `AtomicI32` and implementing the fast-path for read-locking.  Here's a *very* rough sketch:

```rust
fn read_lock(&self) {
loop {
    let x = self.lock.load(Ordering::Relaxed);
    
    if x > 0 {
        if self.compare_exchange_weak(x, x + 1, Ordering::Acquire, Ordering::Relaxed).is_ok() { return ReadGuard; }
        // `x` changed during cmpxchg, retry.
        continue;
    }
    // store into waker set.
    // re-check that the lock wasn't unlocked here and we missed a wakeup.
}
```

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,223 @@
...
+    read_waiters: VecDeque<Waker>,
+    write_waiters: VecDeque<Waker>,
```
hexagonal-sun: Any reason a `WakerSet` wasn't used here?

Also, do we need to distringuish between readers and writers?

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,223 @@
...
+use core::ops::{Deref, DerefMut};
+use core::pin::Pin;
+use core::sync::atomic::{AtomicBool, Ordering};
+use core::task::{Context, Poll, Waker};
+
+use crate::CpuOps;
+
+use super::spinlock::SpinLockIrq;
+
+enum RwlockLockStateInner {
```
hexagonal-sun: lock twice in name?

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,223 @@
...
+                })
+            }
+            Some(RwlockLockStateInner::Read(count)) => {
+                state.is_locked = Some(RwlockLockStateInner::Read(count + 1));
+                Poll::Ready(AsyncRwlockReadGuard {
+                    rwlock: self.rwlock,
+                })
+            }
+            Some(RwlockLockStateInner::Write) => {
+                if state.read_waiters.iter().all(|w| !w.will_wake(cx.waker())) {
```
hexagonal-sun: Any reason we do this check?  I only ask as I don't do this in any other sync primitives, `Mutex` being the main example since it's async.  I suppose it's a matter of weighing up the small likelyhood of a spurious wakeup vs iterating through this list every time the lock is contended?

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,223 @@
...
+                    state.read_waiters.push_back(cx.waker().clone());
```
hexagonal-sun: No drop impl for this future  here so a potential spurious wakeup if the calling future is dropped after we've added ourselves to the waker queue.

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,223 @@
...
+        let mut state = self.rwlock.state.lock_save_irq();
+
+        state.is_locked = None;
+        // Alternate between waking readers and writers to prevent starvation.
+        let was_writer = state
+            .last_woken_was_writer
+            .fetch_update(Ordering::SeqCst, Ordering::SeqCst, |v| Some(!v))
+            .unwrap_or_else(|e| e);
+        if (was_writer && !state.read_waiters.is_empty()) || state.write_waiters.is_empty() {
+            while let Some(next_waker) = state.read_waiters.pop_front() {
```
hexagonal-sun: Why do we wake up everyone here? Won't this create the thundering Hurd problem?

Couldn't we do something like:

If there are writer wakers, wake one.

No writers: therefore wake-up all readers?

This also is why we'd need to distinguish between readers and writers.. so disregard my previous comment on that.

hexagonal-sun (changes requested):

libkernel/src/sync/rwlock.rs
arihant2math: I copied it over from the mutex code, perhaps that has to be changed as well.

Ideally when choosing what to wake, we would wake up all the readers or a single writer.

libkernel/src/sync/rwlock.rs
arihant2math: Essentially I'm trying to be "fair" by alternating between readers and writers when under contention.

When an `Rwlock` is used it's probably because there are more reads than writes, so with that method it would be easy to starve the readers with a continuous stream of write requests.

Now this method isn't 100% fair because it favors readers and writers at 50-50 each, even though it should probably be better proportioned.

I'm chosing either to wake up all the readers or one writer here. All the readers waking up should not be that contentious, especially if we switch to an `AtomicI32` and use fetch_add.

libkernel/src/sync/rwlock.rs
arihant2math: I got this from the mutex code: https://github.com/hexagonal-sun/moss-kernel/blame/96fe0378b7c183aebb4ba27743ba2e9843fcdd8a/libkernel/src/sync/mutex.rs#L96C30-L96C70. I actually have no idea whether we should do this or not, I suppose it is a big performance hit though.

libkernel/src/sync/rwlock.rs
hexagonal-sun: Hm, interesting.  I'll take a look at the mutex code and modify that I think.   I suspect that if a lock is heavily contended, doing this check *every* time is worse than a spurious wake up.

libkernel/src/sync/rwlock.rs
hexagonal-sun: Interesting.  Maybe I need to go back and revisit the `Mutex` code.  That seems like it would be a lot of wasted cycles for contended lock.

libkernel/src/sync/rwlock.rs
hexagonal-sun: Ah I see what you're doing now.  Makes sense.

libkernel/src/sync/rwlock.rs
arihant2math: Ah ok, I see why a `VecDeque` is most efficient. I think for fairness sake we do FIFO, but WakerSet uses a `BTree` internally.

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,216 @@
...
+    rwlock: &'a Rwlock<T, CPU>,
+}
+
+/// A future that resolves to an `AsyncRwlockWriteGuard` when the lock is acquired.
+pub struct RwlockWriteGuardFuture<'a, T: ?Sized, CPU: CpuOps> {
+    rwlock: &'a Rwlock<T, CPU>,
+}
+
+impl<T, CPU: CpuOps> Rwlock<T, CPU> {
+    /// Creates a new asynchronous mutex in an unlocked state.
```
hexagonal-sun: Comment needs updating.

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,216 @@
...
+                lock_state: AtomicI32::new(0),
+                read_waiters: SpinLockIrq::new(VecDeque::new()),
+                write_waiters: SpinLockIrq::new(VecDeque::new()),
+                last_woken_was_writer: AtomicBool::new(false),
+            },
+            data: UnsafeCell::new(data),
+        }
+    }
+
+    /// Consumes the mutex, returning the underlying data.
```
hexagonal-sun: Comment needs updating.

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,216 @@
...
+use super::spinlock::SpinLockIrq;
+
+struct RwlockState<CPU: CpuOps> {
+    lock_state: AtomicI32,
+    read_waiters: SpinLockIrq<VecDeque<Waker>, CPU>,
+    write_waiters: SpinLockIrq<VecDeque<Waker>, CPU>,
+    last_woken_was_writer: AtomicBool,
+}
+
+/// An asynchronous, mutex primitive.
```
hexagonal-sun: Comment needs updating.

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+
+impl<'a, T: ?Sized, CPU: CpuOps> Future for RwlockReadGuardFuture<'a, T, CPU> {
+    type Output = AsyncRwlockReadGuard<'a, T, CPU>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.rwlock.state.lock_state.load(Ordering::Acquire) {
+            0.. => {
+                self.rwlock.state.lock_state.fetch_add(1, Ordering::AcqRel);
```
hexagonal-sun: This code is racy.  What if the lock write-side of the lock is acquired between the `load` and the `fetch_add`?  The trick is to use a CAS operation, that way you know that the initial check on the state of the lock is valid, if the CAS succeeds.  If it fails, restart the whole operation as your initial state is no longer valid.

Rough sketch:

```rust
loop {
    let state = lock_state.load(Ordering::Acquire);

    if state >= 0 {
        if lock_state
            .compare_exchange_weak(
                state,
                state + 1,
                Ordering::AcqRel,
                Ordering::Acquire,
            )
            .is_ok()
        {
            return Poll::Ready(...);
        }
        // Fall thru to the loop to give it another go.
    } else {
       // Locked path.
    }
}
```

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,216 @@
...
+                Poll::Ready(AsyncRwlockReadGuard {
+                    rwlock: self.rwlock,
+                })
+            }
+            -1 => {
+                let mut read_waiters = self.rwlock.state.read_waiters.lock_save_irq();
+                if read_waiters.iter().all(|w| !w.will_wake(cx.waker())) {
+                    read_waiters.push_back(cx.waker().clone());
+                }
```
hexagonal-sun: Likewise here, we need to make sure that the write-side of the lock is still held, otherwise we could be waiting on an unlocked lock.

hexagonal-sun: Note that might be a reason to switch to a `WakerSet`.  When you add yourself in, you obtain a token so you can easily remove yourself from the set using the token.  With a `VecDequeue` that's more difficult since your index in the queue isn't stable.

hexagonal-sun (changes requested):

arihant2math:
Alright, I've decided to switch to a waiter set to reduce the effect of all the queue rechecking and also to help with the de-registering.

arihant2math:
I've decided to rewrite this because I noticed a few other unsafe situations, I think it's safer to use mutexes to represent state rather than re-implementing that, for safety reasons.

arihant2math:
Alright, I was able to massively simplify this by implementing internal state as per https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock#Using_two_mutexes. Should have no soundness issues since I'm not even touching atomics here.

libkernel/src/sync/rwlock.rs
```diff
@@ -0,0 +1,130 @@
...
+        AsyncRwlockReadGuard { rwlock: self }
+    }
+
+    /// Acquires rwlock write.
+    ///
+    /// Returns a future that resolves to a lock guard. The returned future must
+    /// be `.await`ed to acquire the write guard. The guard is released when the
+    /// returned [`AsyncRwlockWriteGuard`] is dropped.
+    pub fn write(&self) -> AsyncRwlockWriteGuard<'_, T, CPU> {
+        self.state.writer_lock.acquire();
```
hexagonal-sun: Doesn't this future just get lost here?  We never await it.

arihant2math: Oops, good catch.

PR #149, opened 2026-01-12
Execute scripts

arihant2math:
Checks for shebang and executes the specified program.


libkernel/src/error.rs
```diff
@@ -111,6 +111,9 @@ pub enum ExecError {
     #[error("Invalid ELF Format")]
     InvalidElfFormat,
 
+    #[error("Invalid Script Format")]
+    InvalidScriptFormat,
```
hexagonal-sun, line 115: Maybe move the error message to `ExecError` enum.

PR #152, opened 2026-01-12
tty: implement termios2 ioctls

hexagonal-sun:
We currently implement `TCGETS`, `TCSETS` and `TCSETSW` ioctls which operate on
a `Termios` struct.  glibc uses the `Termios2` variants of these
syscalls to additionally get the terminal baud rate.

Use a `Termios2` as the main state of the tty.  Return that struct
verbatim  for the newly implemented `Termios2` ioctls.  Convert to/from
the legacy `Termios` struct for the currently implemented ioctls.

This gets bash and `ls` line editing and output working properly under glibc as it uses the newer ioctls.

### DIFF
--- a/src/console/tty.rs
+++ b/src/console/tty.rs
@@ -20,8 +20,8 @@ use libkernel::{
     memory::address::{TUA, UA},
 };
 use meta::{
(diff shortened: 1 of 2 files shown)


arihant2math (approved):
LGTM

PR #155, opened 2026-01-14
Proper timeofday implementations

arihant2math:
Finishes the stub implementation of `get_timeofday` and implements `set_timeofday`. However `set_timeofday` does not handle timezone and `get_timeofday` also ignores it.


hexagonal-sun (approved):
LGTM

PR #160, opened 2026-01-15
Implement `sys_mincore`

arihant2math:
Since we don't page, always returning `0` is valid.


src/memory/mincore.rs
```diff
@@ -0,0 +1,46 @@
...
+    if end <= start {
+        return Err(KernelError::InvalidValue);
+    }
+
+    // Compute number of pages covered by [start, start + len).
+    let pages = len.div_ceil(PAGE_SIZE);
+    if pages == 0 {
+        return Err(KernelError::InvalidValue);
+    }
+
```
hexagonal-sun, line 41: We should also check that `addr` - `addr + len` is a fully contiguous VMA mapped region.  As per the man pages:

```
       ENOMEM addr to addr + length contained unmapped memory.
```

arihant2math:
Alright, I fixed the implementation and actually checked if pages were mapped or not. I actually didn't realize that we having demand paging. Also I've added a sanity test.

src/memory/mincore.rs
```diff
@@ -0,0 +1,74 @@
...
+        .ok_or(KernelError::InvalidValue)?;
+    if end <= start {
+        return Err(KernelError::InvalidValue);
+    }
+
+    // Compute number of pages covered by [start, start + len).
+    let pages = len.div_ceil(PAGE_SIZE);
+    if pages == 0 {
+        return Err(KernelError::InvalidValue);
+    }
```
hexagonal-sun: Consider using `to_mappable_region()` which does the rounding logic:

```rust
    let region = VirtMemoryRegion::new(start_va, len).to_mappable_region().region();
```

src/memory/mincore.rs
```diff
@@ -0,0 +1,74 @@
...
+    let mut buf: Vec<u8> = vec![0; pages];
+
+    {
+        let task = current_task();
+        let mut vm_guard = task.vm.lock_save_irq();
+        let mm = vm_guard.mm_mut();
+
+        // Validate the entire region is covered by VMAs
+        for i in 0..pages {
+            let va = VA::from_value(start_va.value() + i * PAGE_SIZE);
```
hexagonal-sun: Then you can use:

```rust
for (i, page) in range.iter_pages().enumerate() {
    // ...
}
```

src/memory/mincore.rs
```diff
@@ -0,0 +1,74 @@
...
+            if mm.find_vma(va).is_none() {
```
hexagonal-sun, line 44: Isn't this check superfluous?  We've already checked that the entire region is backed by VMAs.

src/memory/mincore.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    {
+        let task = current_task();
+        let mut vm_guard = task.vm.lock_save_irq();
+        let mm = vm_guard.mm_mut();
+
+        let as_ref = mm.address_space_mut();
+
+        for (i, byte) in buf.iter_mut().enumerate() {
+            let va = VA::from_value(start_va.value() + i * PAGE_SIZE);
```
hexagonal-sun: Last point.

```suggestion
            let va = start_va.add_pages(i);
```

hexagonal-sun (approved):
LGTM!

PR #163, opened 2026-01-16
Scheduling updates

### LINKED ISSUE #128: `top` causes a kernel panic when run twice
```
bash-5.3# top
top: can't open 'stat': No such file or directory
bash-5.3# top
[    1.1358589] moss: Kernel panicked at src/sched/current.rs:59:13: Double mutable borrow of current task!
```

arihant2math:
Changes:
- Fix procfs segfault (Fixes #128)
- Schedules tasks on lowest pressure CPUs in SMP, uses a single global atomic for coordinating.

I've tried to safeguard against any contention that might arise from having shared state in a critical path, so I've limited the number of CAS retries to 8, this is fine because it'll retry in ~16 ms anyhow. Also I've limited update times to 16ms for the same purpose, although it might be worth bringing that number down.

### DIFF
--- a/src/sched/mod.rs
+++ b/src/sched/mod.rs
@@ -9,7 +9,7 @@ use crate::{
     process::{TASK_LIST, TaskDescriptor, TaskState},
 };
 use alloc::{boxed::Box, collections::btree_map::BTreeMap, sync::Arc};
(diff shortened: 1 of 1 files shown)


hexagonal-sun (approved):
LGTM

PR #167, opened 2026-01-17
signal: interruptible: prioritize wrapped future over signals

hexagonal-sun:
Poll the wrapped future first, ensuring that the interrupt logic doesn't
short-circuit a future which is already ready.

This fixes an issue where `sys_wait4` would return `-EINTR` when
receiving a `SIGCHLD`. Since `SIGCHLD` indicates the wait condition
is met, the underlying future is ready and should return the PID
successfully rather than aborting.


arihant2math (approved):
LGTM

PR #177, opened 2026-01-24
grep pipe fixes

hexagonal-sun:
- **pipe: add `PipeInode` for fstat**
  Add a PipeInode struct which allows `fstat` for an FD which is backed by
  a pipe.
  

- **devfs: set attr id the same as the inode id**
  Ensure that the ID of the attr contains the ID of the devfs inode,
  instead of the default value.


arihant2math:
huh, same CI failure in #169.

PR #182, opened 2026-01-26
Working busybox `top`

arihant2math:
implements enough of procfs for `top` to run.

### DIFF
--- a/libkernel/src/sync/per_cpu.rs
+++ b/libkernel/src/sync/per_cpu.rs
@@ -332,7 +332,7 @@ mod tests {
 
     #[test]
     fn test_initialization_and_basic_access() {
-        let data: PerCpu<_, MockArch> = PerCpu::new(|| 0u32);
+        let data: PerCpu<_, MockArch> = PerCpu::new(|| RefCell::new(0u32));
         data.init(4); // Simulate a 4-core system
 
         // Act as CPU 0
@@ -355,7 +355,7 @@ mod tests {
     #[test]
     #[should_panic(expected = "PerCpu variable accessed before initialization")]
     fn test_panic_on_uninitialized_access() {
-        let data: PerCpu<_, MockArch> = PerCpu::new(|| 0);
+        let data: PerCpu<_, MockArch> = PerCpu::new(|| RefCell::new(0));
         // This should panic because data.init() was not called.
         let _ = data.borrow();
     }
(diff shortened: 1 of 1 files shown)


libkernel/src/sync/per_cpu.rs
```diff
@@ -130,12 +130,23 @@ impl<T: Send, CPU: CpuOps> PerCpu<T, CPU> {
...
     ///
     /// # Panics
     /// Panics if the `PerCpu` variable has not been initialized.
     fn get_cell(&self) -> &RefCell<T> {
-        let id = CPU::id();
+        let cpu_id = CPU::id();
+        unsafe { self.get_cell_for_cpu(cpu_id) }
+    }
+
+    /// Returns a reference to the underlying data for a different CPU.
```
hexagonal-sun, line 142: My main concern here is that the underlying `RefCell` that wraps `T` will fall over when we have concurrent access across cores.  So, even if `T` is `Send + Sync`, we still will run into problems.

I can  think of two solutions:

1. Split this module into two - `PerCpu` (essentially as it stands right now), and `PerCpuShared`.  We then put a constratit on `T: Send + Sync`, and just have a pointer to an array of `T`s, rather than `RefCell<T>`.
2. Take this module as-is and remove the `RefCell<T>`. This forces the user that wants private-only access to use `RefCell<T>` as their wrapper for `T`.  Then, your new functions can be implemented on a `T` which implements `Send + Sync`:
```rust
impl<T: Send, CPU: CpuOps> PerCpu<T, CPU> {
   // current implementaiton.
}

impl<T: Send + Sync, CPU: CpuOps> PerCpu<T, CPU> {
      fn borrow_for_cpu(&self, cpu_id: usize) -> &T { /* ... */ }
}
```
Notice that we can remove the `unsafe` attribute here, since borrowing across cores is sound when `T: Send + Sync`.

src/process/clone.rs
```diff
@@ -192,6 +194,7 @@ pub async fn sys_clone(
         .insert(tid, Arc::downgrade(&new_task.t_shared));
 
     sched::insert_task_cross_cpu(Box::new(new_task));
+    NUM_FORKS.fetch_add(1, core::sync::atomic::Ordering::SeqCst);
```
hexagonal-sun: Could be `Relaxed` here since we're just using `NUM_FORKS` as a counter, not a barrier.

arihant2math:
I've went with Option 2 and added a bunch of convenience wrappers to the API is basically the same for refcells.

src/sync/per_cpu.rs
```diff
@@ -16,3 +16,23 @@ macro_rules! per_cpu {
         }
     };
 }
+
+/// Wraps with a [`RefCell`] for convenience
+#[macro_export]
+macro_rules! local_per_cpu {
```
hexagonal-sun: Nitpick: maybe call this `per_cpu_private`.

src/sync/per_cpu.rs
hexagonal-sun: Nitpick: maybe call this `per_cpu_shared`.

PR #185, opened 2026-01-28
hardware smp fixes

hexagonal-sun:
- **arm64: memory: flush_to_ram: new**
  Create a new function which takes a pointer to an object `T` and ensures
  that the data is flushed from the caches to the point of
  coherency (RAM).
  

- **arm64: secondary: ensure boot_ctx is flushed to RAM**
  When the secondary boots and derefes the boot_ctx pointer passed in as a
  parameter, it reads NULL, causing the secondary to end up stuck in an
  exception loop. The data is sat in the primary core's cache and hasn't
  been flushed to RAM. Since the secondary starts without the MMU (and
  caches) enabled, the CCI doesn't kick in and we read stale data.
  
  Manually flush the boot_ctx data to RAM before waking up the secondary.
  

- **arm64: boot: secondaries: ensure WFI doesn't trap**
  The WFI insn is set not to trap on the primary core. Ensure it's also
  set on the secondary cores too.

### DIFF
--- a/src/arch/arm64/boot/mod.rs
+++ b/src/arch/arm64/boot/mod.rs
@@ -143,6 +143,9 @@ fn arch_init_secondary(ctx_frame: *mut ExceptionState) -> *mut ExceptionState {
     TCR_EL1.modify(TCR_EL1::EPD0::DisableTTBR0Walks);
     barrier::isb(barrier::SY);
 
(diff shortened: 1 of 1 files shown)


arihant2math:
LGTM after CI

Issue #186, opened 2026-01-29
Why doesn't the global_allocator utilize the FrameAllocator?


zakariamoknine:
Hi, thank you for the amazing work, I've been looking at your memory subsystem with the end goal of constructing my own, and I have a question.

The kernel's global_allocator seems to pre-allocate a fixed 64 MiB region for the heap, why doesn't the global_allocator utilizes the FrameAllocator to shrink/expand itself? pre-allocations can be very punishing for embedded devices where memory is very constraint. For example I have an FPGA with 128 MiB of DDR2, this way, half of the memory will be barely used and not available for userspace.

hexagonal-sun:
Thanks for the kind words and the feedback!

You are absolutely right, pre-allocating 64 MiB is far from ideal, especially for constrained environments like the one you mentioned. The current implementation was a bootstrapping shortcut to get the kernel running before the physical memory management was fully stable.  Since that is now the case, my plan is to replace the current global_allocator with a dynamic allocator that requests pages from the `FrameAllocator` on demand. I am leaning towards a per-CPU slab allocator design; this would solve the memory usage issue and also prevent the lock contention bottleneck in the current single-lock approach.

I am currently finalizing the design for this. If you have any specific thoughts on the implementation, feel free to chime in!

hexagonal-sun:
@zakariamoknine I know you've closed this issue, however I've now written a full slab allocator which means no more pre-allocating a huge chunk of memory for the heap!  It's at #195 if you're interested.

PR #195, opened 2026-02-05
slab allocator

hexagonal-sun:
A working per-cpu caching slab allocator.  

- libkernel: memory: reorganise allocators
Create a new submodule within `memory`, `allocators` which contains all
memory allocators. Also split out the `Frame` struct from the `pg_alloc`
module, allowing it to be used by other modules.

- libkernel: memory: allocs: move FrameList impl into `frame`
Move the FrameList logic out of the physical memory allocator into its
own module. This allows `FrameList` to be shared between the physical
memory allocator and the slab allocator.

- libkernel: memory: PageAllocGetter: return ref to allocator
Rather than returning a reference to the `OnceCell` wrapping the
allocator, return a static reference to the allocator itself. This
allows flexibility for how the allocator is wrapped.

- libkernel: memory: slab: slab: new
Add a new module for managing objects within a slab. The `Slab` struct
manages objects of a given set within a contiguous set of pages. It is
the handle to the underlying memory, allowing for objects to be
allocated, and free'd. It manages a free list of 'indexes' within the
object slots themselves.

- libkernel: memory: slab: allocator: new
Add a new slab allocator implementation. The `SlabAllocator` manages
lists of slabs for each size class.

- libkernel: memory: slab: cache: new
Add a new per-cpu slab cache for caching slab allocations on a given
CPU.

- arch: arm64: exceptions: use `SP_EL0` as scratch reg
When detecting stack overflow in the kernel exception handler, use
`SP_EL0`, rather than `TPIDR_EL1` as the scratch register.  This allows
us to use the `TPIDR_EL1` register as a CPU-banked register for other
purposes.

Use of `SP_EL0` is safe here since the stack overflow check occurs
during kernel exceptions and SP_EL0 will be restored by the user-space
context restore logic, overriding the clobber.

- arch: arm64: memory: heap: new
Add a new heap module which integrates the per-cpu cache and
`SlabAllocator` into a usable `GlobalAlloc` for the kernel.

Fixes: #186

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -239,15 +239,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "linked_list_allocator"
-version = "0.10.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9afa463f5405ee81cdb9cc2baf37e08ec7e4c8209442b5d72c04cfb2cd6e6286"
-dependencies = [
- "spinning_top",
-]
-
-[[package]]
 name = "lock_api"
 version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -300,7 +291,6 @@ dependencies = [
  "futures",
  "getargs",
  "libkernel",
- "linked_list_allocator",
  "log",
  "object",
  "paste",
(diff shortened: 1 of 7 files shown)


arihant2math (approved):
LGTM

PR #196, opened 2026-02-05
Mount syscall

arihant2math:
Implements `mount`. Includes a procfs bugfix because systemd mounts via procfs symlinks.

Testing: systemd

### DIFF
--- a/src/fs/syscalls/mount.rs
+++ b/src/fs/syscalls/mount.rs
@@ -1,6 +1,6 @@
 use crate::fs::VFS;
 use crate::memory::uaccess::cstr::UserCStr;
-use crate::sched::current::current_task_shared;
+use crate::sched::syscall_ctx::ProcessCtx;
 use bitflags::bitflags;
 use core::ffi::c_char;
 use libkernel::error::{KernelError, Result};
@@ -46,6 +46,7 @@ bitflags! {
 }
 
 pub async fn sys_mount(
+    ctx: &ProcessCtx,
     dev_name: TUA<c_char>,
     dir_name: TUA<c_char>,
     type_: TUA<c_char>,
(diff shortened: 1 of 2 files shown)


arihant2math:
Seems like it is possible for this implementation to fault in normal operation, still debugging.

arihant2math:
Ah, `dev_name` and `type` are optional.

src/fs/syscalls/mount.rs
```diff
@@ -0,0 +1,93 @@
...
+    let flags = MountFlags::from_bits_truncate(flags as u64);
+    if flags.contains(MountFlags::MS_REC) {
+        // TODO: Handle later
+        return Ok(0);
+    }
+    let mut buf = [0u8; 1024];
+    let dev_name = UserCStr::from_ptr(dev_name)
+        .copy_from_user(&mut buf)
+        .await
+        .ok();
```
hexagonal-sun: Couldn't this be `.await?`?

src/fs/syscalls/mount.rs
```diff
@@ -0,0 +1,93 @@
...
+        const MS_BORN = 1 << 29;
+        const MS_ACTIVE	= 1 << 30;
+        const MS_NOUSER	= 1 << 31;
+    }
+}
+
+pub async fn sys_mount(
+    dev_name: TUA<c_char>,
+    dir_name: TUA<c_char>,
+    type_: TUA<c_char>,
```
hexagonal-sun, line 52: I'm a little confused by the args.  According to the man page, `mount` looks like:

```c
       int mount(const char *source, const char *target,
                 const char *filesystemtype, unsigned long mountflags,
                 const void *_Nullable data);

```

But the following code looks as though it's inspecting the `source` argument for the filesystem type.  Shouldn't that be `type`?

arihant2math: From the table in `etc/` it says:
```
(char *dev_name, char *dir_name, char *type, unsigned long flags, void *data)
```

`filesystemtype` (or `type`) is `null` when `systemd` calls this (which was what led to the `EFAULT` errors). I've got no clue why systemd calls this API the way does. Perhaps `source` could also be a device file, in which case `type` is used to specify the filesystem type?

hexagonal-sun: Hm, odd. I can understand why `type` could be null, you'd read the FS type from `/etc/fstab`, or inspect the block device to see if any of your FS driver match the magic and try to mount.  Odd that it's not specified as nullable in the man page through.

Issue #190, opened 2026-01-30
Systemd blockers


arihant2math:
- [x] More prctl #153 
- [x] `sys_close_range` #154 
- [x] `sys_waitid` #166 
- [x] `sys_clock_settime` #219
- [x] `sys_statfs` and `sys_fstatfs` #203
- [x] `sys_mount` #196 
- [x] `sys_flock` (stubbed)
- [x] `/proc/cmdline` #189 
- [x] `/dev/shm`
- [x] `/proc/<pid>/fd/<fd>` #222
- [x] sysfs #197  
- [x] cgroup filesystem #197 (good enough for now)
- [x] `sys_pidfd_open`
- [x] `sys_sethostname` #201 (fixup not merged)
- [x] `sys_copy_file_range` #218 
- [x] epoll #282 
- [x] `sys_signalfd4` #282
- [ ] more cgroup #285 
- [ ] inotify
- [x] socket support #39

arihant2math:
Currently systemd is failing in the mounting step.

PR #201, opened 2026-02-08
Store hostname and allow for updates via `sys_sethostname`

arihant2math:
Part of #190.

### DIFF
--- a/src/kernel/hostname.rs
+++ b/src/kernel/hostname.rs
@@ -1,8 +1,8 @@
-use crate::memory::uaccess::copy_from_user_slice;
+use crate::memory::uaccess::cstr::UserCStr;
 use crate::sched::current::current_task_shared;
 use crate::sync::OnceLock;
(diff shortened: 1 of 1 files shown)


src/kernel/hostname.rs
```diff
@@ -0,0 +1,49 @@
...
+        creds
+            .caps()
+            .check_capable(CapabilitiesFlags::CAP_SYS_ADMIN)?;
+    }
+
+    if name_len > HOST_NAME_MAX {
+        return Err(KernelError::NameTooLong);
+    }
+    let mut buf = vec![0u8; name_len];
+    copy_from_user_slice(name_ptr.to_untyped(), &mut buf).await?;
```
hexagonal-sun: There are specific functions for copying c-strings from userspace.  Use something like:

```rust
        let str = UserCStr::from_ptr(ptr).copy_from_user(&mut buf).await?;
```

PR #202, opened 2026-02-09
libkernel: memory: slab: heap: new

hexagonal-sun:
Add a new module which implements the high-level heap logic (implements
`GlobalAlloc`) for the slab allocator.

Also implement a multi-threaded stress test of the heap ensuring
concurrent allocation is unique (doesn't corrupt other allocs) and all
memory is free'd back to the `FrameAllocator` when all caches are
purged.


arihant2math (approved):
LGTM

PR #207, opened 2026-02-11
Initial epoll implementation

arihant2math:
The usertest still fails due to bug in the `sys_epoll_pwait` implementation, but everything else is fine.


arihant2math:
Closing until substantive

PR #209, opened 2026-02-12
Update ext4-view fork

arihant2math:
- Reduces per-inode memory usage
- Updates API to rely less on the metadata structure (will be removed eventually)

Internally the inode does not track metadata as a structure separately as a field, since it duplicates the information already present in it's internal buffer. Now it reads the buffer on demand to generate the struct.

Additionally a lot of scaffolding was laid internally, but that doesn't affect any of the interfaces that moss uses currently.


hexagonal-sun (approved):
LGTM

PR #216, opened 2026-02-15
Fix macos scripts

andrewdavidmackenzie:
Avoid scripts not terminating, but asking user for commands from standard input.
Documented a couple of dependency issues I ran into on macos to build.


andrewdavidmackenzie:
I had to install zig also, although I didn't do anything special to build musl....not sure if that is expected. 
If always needed, maybe should be moved to the dependencies section above?

arihant2math:
Yes, zig is required, and it comes built in with cross compilation toolchains.

scripts/mac-experimental/README.md
arihant2math, line 22: Although, that is made note of here.

arihant2math (approved):

Issue #215, opened 2026-02-15
macos building problem with debugfs


andrewdavidmackenzie:
I would like to get the build working on macos, as that's the machine I work on the most.
I cloned the repo and tried the scripts, following the readme in the mac-experimental directory:
- `build-deps.sh` seems to work fine
- `create-image.sh` leaves me in debugfs interactively, and I'm not sure what command to type or how to exit correctly. If I exit with Control-D, no image file is created.

So, I suspect something is different on macos relaetd to debugfs that makes this happen.

Investigating I find:
" When debugfs is run with -f file, it processes the commands from the file but then
   continues to read from stdin for additional commands. This is why it appears to hang - it's waiting 
  for more input.   "

andrewdavidmackenzie:
Preparing a PR that fixes that issue.
It builds but running fails with `0.021576] moss: Kernel panicked at src/main.rs:128:10: Unable to find init: Fs(NotFound)` , which maybe unrelated

hexagonal-sun:
I'm afraid that I don't run MacOS so I can't help here. I know @arihant2math does, any ideas?

arihant2math:
@hexagonal-sun I've stopped using the build scripts and just use an alpine rootfs, it might be worth just switching over to that since it's way simpler. Image creation turns into:
```
rm moss.img
mkfs.ext4 moss.img 512MB
mkfs.ext4 -d build/alpine moss.img
```

across macos and linux.

I'm honestly not sure if `debugfs` works correctly.

arihant2math:
#217 should fix this issue entirely.
For now, you can download `moss.img` from https://github.com/hexagonal-sun/moss-kernel/actions/runs/22042194841 and place it in the project root. That should have everything working as expected.

andrewdavidmackenzie:
Fails for me with:

```
 mkfs.ext4 moss.img 512MB
mke2fs 1.47.3 (8-Jul-2025)
mkfs.ext4: invalid blocks '512MB' on device 'moss.img'
```

Maybe I'll just wait for the build system revamp in #217 if you are going to merge it soon.

If the PR is ready to merge, let me know and I can try it on macos for you.

andrewdavidmackenzie:
This looks like it works with the downloaded image:
```
amackenz@amackenz-mac moss-kernel % cargo run --release
    Finished `release` profile [optimized] target(s) in 0.04s
warning: the following packages contain code that will be rejected by a future version of Rust: memchr v2.7.6
[...]

PR #217, opened 2026-02-15
Update the build system

arihant2math:
Since the build system has been written, we've added support for dynamic linking. This allows us to load an alpine rootfs image without modification. This reduces image creation time and increases compatibility because everything comes in the box. The only addition is usertest, which I copy over in the script.

- Use alpine for default image instead of the custom build
- Uses `just` to run commands (although they generally one-liners and you can do without installing it)
- macOS isn't experimental anymore!
- Require `jq` (needed to find where the usertest compiled binary is)

Fixes: #215


andrewdavidmackenzie:
On macos:
```
amackenz@amackenz-mac moss-kernel % just run         
# Create moss.img if it doesn't exist
if [ ! -f moss.img ]; then
sh: -c: line 1: syntax error: unexpected end of file
error: Recipe `run` failed on line 8 with exit code 2
```

running commands separately works:
```
# First time only (to create the image)
./scripts/create-image.sh
# Then, to run the kernel in QEMU
# By default it will launch into bash, which alpine doesn't have, however `ash` and `sh` are both available.
cargo run --release -- /bin/ash
```

Claude helped me out with `just` to make the multi-line shell script at line 7 work

```
run:
    #!/usr/bin/env sh
    # Create moss.img if it doesn't exist
    if [ ! -f moss.img ]; then
        just create-image
    fi
    cargo run --release -- /bin/ash
```

and now `just run` works.

Justfile
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env just --justfile
+
+create-image:
+    ./scripts/create-image.sh
+
+run:
+    # Create moss.img if it doesn't exist
+    if [ ! -f moss.img ]; then
```
andrewdavidmackenzie, line 9: Modify to 
```
run:
    #!/usr/bin/env sh
    # Create moss.img if it doesn't exist
    if [ ! -f moss.img ]; then
        just create-image
    fi
    cargo run --release -- /bin/ash
```

to make a multi-line script and avoid a failure running it (for me, on macos, with `just 1.46.0`

arihant2math: Good catch!

andrewdavidmackenzie:
I have confirmed that (with a minor change that I have commented in the PR) this builds and runs successfully on macos (Tahoe 26.3)

arihant2math:
Since I do have a minor use case that wasn't being supported, I've added `build/extra_bins` for other binaries that will get copied over to `/bin` in the rootfs. Useful for things like strace.

hexagonal-sun (approved):
LGTM, I'll give it a go on Linux and if all works okay, I'll merge.

arihant2math:
Whoops ... I'll revert this.

PR #200, opened 2026-02-08
Systemd fixups

arihant2math:
Probably will run into 1) needing a r/w filesystem and 2) socket implementation sometime soon.

I'll be backporting fixes from this branch over to master in subprs.

PR #218, opened 2026-02-16
Implement `copy_file_range` syscall

arihant2math:
Part of #200.

[man copy_file_range](https://man7.org/linux/man-pages/man2/copy_file_range.2.html)


src/fs/syscalls/copy_file_range.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            remaining -= read;
+        }
+
+        return Ok(total_written);
+    }
+
+    // Offset path: at least one of the offsets was provided. Use readat/writeat
+    // with a kernel buffer and manual offset maintenance.
+    let mut in_off: u64 = if off_in.is_null() {
```
hexagonal-sun: If the user passes in `off_in` and `off_out` as valid pointers but they point to a value of `0`, we could still use the fast path.  Maybe try to determine the offsets at the start of the fn, and then the check on line 43 could be a check for `== 0`, rather than null to catch this case.

src/fs/syscalls/copy_file_range.rs
```diff
@@ -0,0 +1,255 @@
...
+        if v < 0 {
+            return Err(KernelError::InvalidValue);
+        }
+        v as u64
+    };
+
+    // We'll use a scratch page as the backing store for a kernel pipe buffer.
+    // This keeps the implementation simple and avoids user buffers.
+    let mut pg = ClaimedPage::alloc_zeroed()?;
+    let kslice = pg.as_slice_mut();
```
hexagonal-sun: I don't think this is needed since `KPipe` alloocates it's own page on `::new()`.

hexagonal-sun (approved):
LGTM

PR #227, opened 2026-02-20
Rework uname to avoid hardcoded datetime and add unit tests

andrewdavidmackenzie:
A PR to get my hands dirty with moss-kernel, more than be tremendously useful.

Restructure `uname` so that:
- there is an internal function that does the main logic and can be tested by unit tests
- Generate an env var in build.rs that is the actual build datetime timestamp
- Avoid hardcoded date in the version string
- Use the timestamp in the building of `uname` response, so the data is the actual build time
- add a simple unit test for the sysname field
- add more complex test for the version field that checks the date formatting (not exhaustively)

NOTE: The build timestamp can be used by any other function, syscall or part of the code if required.

TODO: Clarify the use of the "https://github.com/andrewdavidmackenzie/moss-kernel/pull/1" at the start and if that is a build number that wants to be incremented somehow (between releases?)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,6 @@
 use std::path::PathBuf;
-use time::macros::format_description;
 use time::OffsetDateTime;
+use time::macros::format_description;
 
 fn main() {
     let linker_script = match std::env::var("CARGO_CFG_TARGET_ARCH") {


Cargo.toml
```diff
@@ -32,6 +32,9 @@ futures = { version = "0.3.31", default-features = false, features = ["alloc", "
 rand = { version = "0.9.2", default-features = false, features = ["small_rng"] }
 rustc-hash = { version = "2.1", default-features = false }
 
+[build-dependencies]
+chrono = "0.4.43" # For build timestamping vis build.rs
```
arihant2math: Might be worth using `time` instead; it's a bit more lightweight and should reduce the build time a little.

andrewdavidmackenzie: I can try it. Never used it. Taking a look at it's source, it's not light on sub-dependencies either...

https://github.com/time-rs/time/blob/main/Cargo.toml

vs.

https://github.com/chronotope/chrono/blob/main/Cargo.toml

arihant2math: time seems to be lighter on the whole because it doesn't support timezone localization related stuff I think? I'll try to disable as many of the dependencies as I can.

andrewdavidmackenzie:
I usually do default-features=false, then add features back in until it
works :-)

PR #228, opened 2026-02-20
Make `ktest` a proc macro and remove docker from CI

arihant2math:
- Add command to `just` for running kunit tests.
- Remove docker building from CI to decrease CI time (by around ~50%!). Image creation has gotten far easier, so the overhead of docker containers can be cut by just executing directly on the CI. This also helps us benefit from CI caching of dependencies.


hexagonal-sun (approved):
LGTM!

Issue #220, opened 2026-02-16
Intermittency in `heap_stress_test`


arihant2math:
Source: https://github.com/hexagonal-sun/moss-kernel/actions/runs/22051684554/job/63710990046?pr=218

```
---- memory::allocators::slab::heap::tests::heap_stress_test stdout ----
Initial Free Pages: 129024

thread '<unnamed>' (3686) panicked at libkernel/src/memory/allocators/slab/allocator.rs:217:59:
Should have free slabs
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: <core::option::Option<intrusive_collections::unsafe_ref::UnsafeRef<libkernel::memory::allocators::frame::Frame>>>::expect
   4: <libkernel::memory::allocators::slab::allocator::SlabManager<libkernel::test::MockCpuOps, libkernel::memory::allocators::slab::heap::tests::TestAllocGetter, libkernel::memory::address::IdentityTranslator>>::free
   5: <libkernel::memory::allocators::slab::cache::PtrCache>::drain_into::<libkernel::test::MockCpuOps, libkernel::memory::allocators::slab::heap::tests::TestAllocGetter, libkernel::memory::address::IdentityTranslator>
   6: <libkernel::memory::allocators::slab::heap::KHeap<libkernel::test::MockCpuOps, libkernel::memory::allocators::slab::heap::tests::ThreadLocalCacheStorage, libkernel::memory::allocators::slab::heap::tests::TestAllocGetter, libkernel::memory::address::IdentityTranslator, libkernel::memory::allocators::slab::heap::tests::TestSlabGetter> as core::alloc::global::GlobalAlloc>::dealloc
   7: libkernel::memory::allocators::slab::heap::tests::heap_stress_test::{closure#0}
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

thread 'memory::allocators::slab::heap::tests::heap_stress_test' (3656) panicked at libkernel/src/memory/allocators/slab/heap.rs:377:22:
called `Result::unwrap()` on an `Err` value: Any { .. }
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: libkernel::memory::allocators::slab::heap::tests::heap_stress_test
[...]

PR #229, opened 2026-02-21
libkernel/slab: fix UB and accounting bugs identified by Miri

hexagonal-sun:
Fix several issues in the slab allocator identified by the heap stress
tests (intermittent) and running the tests under Miri.

1. Fix Undefined Behavior. Previously, the allocator created temporary
   `&mut Frame` references while raw pointers to that frame existed in the
   intrusive `partial` or `free` lists. Under Miri's strict aliasing rules,
   creating a unique reference to the whole struct invalidated the list
   pointers.

   The fix implements "split borrowing": we now maintain raw pointers to
   the frame and only create mutable references to the `.state` field
   when needed, ensuring the `.link` field remains valid for the
   intrusive list.

2. Fix `free_list_sz` accounting. In `try_alloc`, the `free_list_sz`
   counter was not being decremented when a slab was successfully popped
   from the free list. This caused the allocator to believe it had free
   slabs when the list was actually empty, leading to panics during
   batch-free operations.

3. Increase heap stress tests. The test suite now runs the stress test
   in a loop to catch state persistence bugs and ensures the allocator is
   fully torn down and reset between iterations.

Fixes: #220

### DIFF
--- a/libkernel/src/memory/allocators/frame.rs
+++ b/libkernel/src/memory/allocators/frame.rs
@@ -37,7 +37,8 @@ pub enum FrameState {
 #[derive(Debug, Clone)]
 pub struct Frame {
     pub state: FrameState,
-    pub link: LinkedListLink, // only used in free nodes.
+    // used in free nodes list (FA), Free and Partial Lists (SA).
+    pub link: LinkedListLink,
     pub pfn: PageFrame,
 }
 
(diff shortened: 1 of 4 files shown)


arihant2math (approved):
LGTM for these changes, I'll do a more comprehensive review later

PR #232, opened 2026-02-23
Bump dependencies + more ext4 writing

arihant2math:
- rand `0.9 -> 0.10`
- intrusive_collections `0.9 -> 0.10`
- ext4_view

## ext4 writing

- Adds support for extendable files (but not contractable) via extents (no block map support yet).
- File deletion (frees all blocks and inode)
- File creation
- Add/remove items from directories (including hardlinks and subdirectories) (non-htree)
- Symlink support (both for "fast symlinks" and regular symlinks)
- New directory creation (non-htree)
- Some utility methods for the time being (`rename_from` can be optimized a little)
- and partridge and a pear tree

Currently `apk` can install `libunistring` via `apk add --repositories-file=repo.list --allow-untrusted --no-network --no-cache libunistring-1.4.1-r0.apk` (create a blank `repo.list` first with `touch repo.list`)

also I switched to workspace dependencies where possible to minimize random cases of dependency mismatch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,7 +142,7 @@ dependencies = [
 [[package]]
 name = "ext4-view"
 version = "0.9.3"
-source = "git+https://github.com/arihant2math/ext4-view-rs.git?branch=main#c95f2de083f50c2d04e370048ac6d2a9fbe8c516"
+source = "git+https://github.com/arihant2math/ext4-view-rs.git?branch=main#cb9a222d37f7b90d3b92068cad8d86d042171471"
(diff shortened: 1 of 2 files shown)


hexagonal-sun:
Are there issues with these uplifts?  Looks like it passes CI.

arihant2math:
> Are there issues with these uplifts? Looks like it passes CI.

Just fixed a couple of panics with hardlink creation/deletion, should be good now.

hexagonal-sun (approved):
LGTM!

PR #239, opened 2026-03-01
Add `find_process_by_tgid`

arihant2math:
I was using this when implementing itimers and realized the same pattern was being used elsewhere.


hexagonal-sun (approved):
Nice catch

PR #240, opened 2026-03-01
Don't force schedule every 50ms

arihant2math:


arihant2math:
So there seems to be a bug in the preemption scheduling logic:
https://github.com/hexagonal-sun/moss-kernel/actions/runs/22534060821/job/65278145288?pr=240
https://github.com/hexagonal-sun/moss-kernel/actions/runs/22534060821/job/65278145282?pr=240

PR #241, opened 2026-03-01
Fix double borrow of task when delivering user signals

TroyNeubauer:
While experimenting with #237, I found a double borrow:
```
moss: Kernel panicked at src/arch/arm64/memory/fault.rs:46:20: Double mutable borrow of current task! Borrowed from: Some(Location { file: "src/arch/arm64/proc/signal.rs", line: 30, column: 16 })
```
Borrowing current_task over the [copy_to_user](https://github.com/hexagonal-sun/moss-kernel/blob/912dc0f713a6741f27876d3aaac772749b36668e/src/arch/arm64/proc/signal.rs#L59) in [do_signal](https://github.com/hexagonal-sun/moss-kernel/blob/912dc0f713a6741f27876d3aaac772749b36668e/src/arch/arm64/proc/signal.rs#L30) causes a double borrow when [run_mem_fault_handler](https://github.com/hexagonal-sun/moss-kernel/blob/912dc0f713a6741f27876d3aaac772749b36668e/src/arch/arm64/memory/fault.rs#L46) runs (since the user stack is demand-paged)

It would be nice to have the type system help us here, but looks like `CurrentTaskGuard` is already `!Send` + `!Sync`.

### DIFF
--- a/src/sched/current.rs
+++ b/src/sched/current.rs
@@ -17,7 +17,7 @@ per_cpu_private! {
 pub(super) struct CurrentTaskPtr {
     pub(super) ptr: Cell<*mut OwnedTask>,
     pub(super) borrowed: Cell<bool>,
-    pub location: Cell<Option<core::panic::Location<'static>>>,
+    location: Cell<Option<core::panic::Location<'static>>>,
 }
 
 unsafe impl Send for CurrentTaskPtr {}


hexagonal-sun (approved):
Nice catch!

I presume the segfault tests you've added to usertest tripped the double borrow?

TroyNeubauer:
Yeah - if you comment `drop(signal); drop(task);` you will trigger the double borrow

hexagonal-sun:
Merged. Thanks for the contribution!

PR #242, opened 2026-03-02
Initial display support

arihant2math:
Exposes primary display via `/dev/fb0`. Doesn't support any ioctl and panics on selected ones to prevent faults. An easy way to test this is `cat /dev/random >> /dev/fb0`.

Also implements an virtio HAL for use in other virtio drivers.

To enable the display, use `--display` with `cargo run`.

### DIFF
--- a/src/drivers/display/mod.rs
+++ b/src/drivers/display/mod.rs
@@ -11,16 +11,14 @@ use alloc::{boxed::Box, sync::Arc};
 use async_trait::async_trait;
 use core::pin::Pin;
 use libkernel::driver::CharDevDescriptor;
-use libkernel::error::KernelError;
+use libkernel::error::{FsError, KernelError};
 use libkernel::fs::OpenFlags;
 use libkernel::fs::attr::FilePermissions;
 use libkernel::memory::address::UA;
 
 pub mod virtio;
 
-/// Minimal kernel-facing display abstraction.
-///
-/// This is intentionally small: it provides a framebuffer (RGBA8888) and the
+/// Kernel display abstraction: a framebuffer (RGBA8888) and the
 /// ability to flush updated contents to the device.
 ///
 /// The underlying implementation may be MMIO, VirtIO, etc.
@@ -185,6 +183,9 @@ struct FbDev;
 
 impl OpenableDevice for FbDev {
     fn open(&self, flags: OpenFlags) -> libkernel::error::Result<Arc<OpenFile>> {
+        if system_display().is_none() {
+            Err(FsError::NotFound)?;
+        }
         Ok(Arc::new(OpenFile::new(Box::new(FbFileOps), flags)))
     }
 }
--- a/src/drivers/virtio_hal.rs
+++ b/src/drivers/virtio_hal.rs
@@ -1,4 +1,5 @@
 use crate::arch::ArchImpl;
+use crate::memory::PageOffsetTranslator;
 use core::ptr::NonNull;
 use libkernel::VirtualMemory;
 use libkernel::memory::PAGE_SIZE;
@@ -6,7 +7,6 @@ use libkernel::memory::address::{PA, TPA};
 use libkernel::memory::region::PhysMemoryRegion;
 use log::trace;
 use virtio_drivers::{BufferDirection, Hal, PhysAddr};
-use crate::memory::PageOffsetTranslator;
 
 pub(super) struct VirtioHal;
 
@@ -67,7 +67,9 @@ unsafe impl Hal for VirtioHal {
     }
 
     unsafe fn mmio_phys_to_virt(paddr: PhysAddr, _size: usize) -> NonNull<u8> {
-        let vaddr = TPA::from_value(paddr as usize).to_va::<PageOffsetTranslator>().as_ptr_mut();
+        let vaddr = TPA::from_value(paddr as usize)
+            .to_va::<PageOffsetTranslator>()
+            .as_ptr_mut();
         NonNull::new(vaddr).unwrap()
     }
 


src/drivers/virtio_hal.rs
```diff
@@ -0,0 +1,96 @@
...
+pub(super) struct VirtioHal;
+
+impl VirtioHal {
+    #[inline]
+    fn pages_to_order(pages: usize) -> u8 {
+        // virtio-drivers asks in pages; our physical allocator takes a power-of-two order.
+        // Round up to the next power-of-two to ensure we have enough contiguous frames.
+        let pages = pages.max(1);
+        let rounded = pages.next_power_of_two();
+        (usize::BITS - 1 - rounded.leading_zeros()) as u8
```
hexagonal-sun: Couldn't you use `ilog2()` function here to convert to order?

src/drivers/virtio_hal.rs
```diff
@@ -0,0 +1,96 @@
...
+            .expect("PAGE_ALLOC not initialized")
+            .alloc_frames(order)
+            .expect("virtio dma_alloc: out of memory");
+
+        let region_start = alloc.region().start_address();
+        let paddr = region_start.value() as PhysAddr;
+
+        // Leak so the pages remain owned until `dma_dealloc`.
+        let leaked = alloc.leak();
+        debug_assert_eq!(leaked.start_address().value(), region_start.value());
```
hexagonal-sun: I'm not sure this is needed.  I'd just use the region returned by `.leak()`:

```rust
let region = crate::memory::PAGE_ALLOC
            .get()
            .expect("PAGE_ALLOC not initialized")
            .alloc_frames(order)
            .expect("virtio dma_alloc: out of memory")
             .leak();
```

src/drivers/virtio_hal.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        // Convert PA->VA using the kernel's direct mapping window.
+        let vaddr = (ArchImpl::PAGE_OFFSET + region_start.value()) as *mut u8;
```
hexagonal-sun: So, you should be able to do:

```rust
let vaddr = region.start_addrsss().to_va::<PageOffsetTranslator>()
```

You can also translate the whole region to a `VirtMemoryRegion` with `to_va()` too.

src/drivers/virtio_hal.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        drop(alloc);
+        0
+    }
+
+    unsafe fn mmio_phys_to_virt(paddr: PhysAddr, _size: usize) -> NonNull<u8> {
+        // Most callers in this repo map MMIO explicitly before constructing the transport,
+        // but avoid assuming identity mapping here.
+        // For now, rely on the direct map if the address is in RAM; otherwise this is a bug.
+        let vaddr = (ArchImpl::PAGE_OFFSET + (paddr as usize)) as *mut u8;
```
hexagonal-sun: I'd use the `Address` types here:

```rust
PA::from_value(paddr).to_va::<PageOffsetTranslator>().as_ptr_mut()
```

src/drivers/virtio_hal.rs
```diff
@@ -0,0 +1,96 @@
...
+        NonNull::new(vaddr).unwrap()
+    }
+
+    unsafe fn share(buffer: NonNull<[u8]>, _direction: BufferDirection) -> PhysAddr {
```
hexagonal-sun, line 76: It's probably worth noting here that we're assuming that all ram is DMA-coherent in QEMU.  We don't need to adjust page table mapping properties to disable caching and the like to make this work.

src/arch/mod.rs
```diff
@@ -195,6 +195,11 @@ pub trait Arch: CpuOps + VirtualMemory {
         dst: *mut u8,
         len: usize,
     ) -> impl Future<Output = Result<usize>>;
+
+    /// Translates a kernel virtual address to a physical address. This is used for
+    /// mapping kernel symbols into the MMIO region for userspace access, and for
+    /// translating kernel addresses to physical addresses for DMA.
+    fn translate_kernel_va(addr: VA) -> PA;
```
hexagonal-sun: Not too sure why this is needed?  I can't see it used in any of the other code.

src/arch/arm64/mod.rs
```diff
@@ -172,4 +173,8 @@ impl Arch for Aarch64 {
     ) -> impl Future<Output = Result<usize>> {
         Arm64CopyStrnFromUser::new(src, dst as *mut _, len)
     }
+
+    fn translate_kernel_va(addr: VA) -> PA {
```
hexagonal-sun: Likewise, I can't see where this is used.

src/drivers/display/virtio.rs
```diff
@@ -0,0 +1,176 @@
...
+            let transport = unsafe {
+                MmioTransport::new(header, size).map_err(|e| {
+                    log::error!("Failed to initialize virtio-mmio transport: {e}");
+                    KernelError::Other("virtio-mmio transport init failed")
+                })?
+            };
+
+            // Only bind to GPU here; other virtio-mmio devices should be handled by their own drivers.
+            if !matches!(transport.device_type(), DeviceType::GPU) {
+                return Err(KernelError::Probe(ProbeError::Deferred));
```
hexagonal-sun, line 138: I really should come up with a new ProbeError type here.  Maybe something like `CannotHandle`, or something indicating that the probe was successful, but for whatever reason this driver cannot handle the device.

Another option would be for this funciton to return `Result<Option<Arc<dyn Driver>>>`, but I'm not sure if I'm such a fan of that.

src/drivers/display/virtio.rs
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            if mapped != VA::from_value(0xffffd00000f92e00) {
```
hexagonal-sun, line 141: Random address?  Is this debug code?

src/drivers/display/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// Creates a framebuffer guard from an existing lock guard plus a framebuffer slice.
+    ///
+    /// This is primarily intended for driver implementations.
+    pub fn new<G>(guard: G, fb: &'a mut [u8]) -> Self
+    where
+        G: 'a,
+    {
+        Self {
+            fb,
+            _drop_guard: Box::new(move || drop(guard)),
```
hexagonal-sun, line 60: Not too sure why this needs to be in a closure just to call drop here?  Can't the inner guard just be a member of `FrameBufferGuard<'a, G>`?  Then when the guard is dropp it'll also drop the inner guard automatically?

src/drivers/display/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    /// Flush any pending framebuffer updates to the physical display.
+    fn flush(&self) -> libkernel::error::Result<()>;
+}
+
+/// A guard that keeps the underlying display locked while the framebuffer slice is borrowed.
+///
+/// Internally, this stores a small drop-closure that owns the lock guard. When the
+/// `FramebufferGuard` is dropped, the closure is dropped, releasing the lock guard.
+pub struct FramebufferGuard<'a> {
```
hexagonal-sun, line 45: Not too sure what this is meant for?   Are you trying to implement an intrusive lock here?

hexagonal-sun (changes requested):

src/drivers/display/virtio.rs
arihant2math: For some reason, the frame buffer at that address was causing crashes, I'll debug that first.

src/drivers/display/mod.rs
arihant2math: The framebuffer itself is just a slice of memory, so I use this to lock it. The guard is the Spinlock for the gpu attribute in the virtio implementation.

src/drivers/display/mod.rs
arihant2math: Due to the `trait Display` having `fn lock_framebuffer(&self) -> FramebufferGuard<'_>`, I'd have to add a generic to that as well, but I can't since I need arbitrary displays for `SYS_DISPLAY` and associated methods

PR #246, opened 2026-03-05
virtio rng

hexagonal-sun:
- **Add BLAKE2s entropy pool with per-CPU ChaCha20 CSPRNG**
  Replace the single `SmallRng` with a proper two-layer RNG architecture:
  
  - Global BLAKE2s-256 entropy pool that accumulates entropy and gates
    seed extraction behind a 256-bit threshold.
  - Per-CPU ChaCha20Rng instances that are lazily seeded from the pool on
    first use and periodically reseed every 1 MB by XOR-ing a fresh BLAKE2
    extract with 32 bytes of their own output.
  
  The /dev/random chardev uses fill_random_bytes directly instead of
  routing through the syscall layer.
  

- **tty: add entropy**
  Add entropy to the entropy pool when processing input on TTYs.
  

- **rand: add `EntropySource`**
  Add an `EntropySource` trait which allows the `EntropyPool` to pull
  entropy from various sources.
  

- **Add virtio-rng driver and ProbeError::NoMatch for silent probe skipping**
  Add a virtio-rng entropy source driver that registers with the kernel
  RNG subsystem. Introduce `ProbeError::NoMatch` so virtio drivers can
  silently reject empty or wrong-type MMIO slots without spamming the boot
  log with fatal errors or leaving devices stuck in the deferred queue.

### DIFF
--- a/scripts/qemu_runner.py
+++ b/scripts/qemu_runner.py
@@ -41,8 +41,11 @@ default_args = {
     "-nographic": None,
     "-s": None,
     "-kernel": bin_executable_location,
-    "-append": f"{append_args} --rootfs=ext4fs --automount=/dev,devfs --automount=/tmp,tmpfs --automount=/proc,procfs --automount=/sys,sysfs"
+    "-append": f"{append_args} --rootfs=ext4fs --automount=/dev,devfs --automount=/tmp,tmpfs --automount=/proc,procfs --automount=/sys,sysfs",
 }
+
+# Arguments that can appear multiple times (e.g. -device)
+extra_args = ["-device", "virtio-rng-device"]
 
 if args.debug:
     default_args["-S"] = None
@@ -53,7 +56,7 @@ if args.display:
     default_args["-nic"] = "none"
     # Add uart
     default_args["-serial"] = "stdio"
-    default_args["-device"] = "virtio-gpu-device"
+    extra_args += ["-device", "virtio-gpu-device"]
 
 qemu_command = ["qemu-system-aarch64"]
 
@@ -61,5 +64,7 @@ for key, value in default_args.items():
     qemu_command.append(key)
     if value is not None:
         qemu_command.append(value)
+
+qemu_command += extra_args
 
 subprocess.run(qemu_command, check=True)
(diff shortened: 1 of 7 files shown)


scripts/qemu_runner.py
```diff
@@ -23,7 +23,7 @@
 elf_executable = args.elf_executable
 bin_executable_location = elf_executable.replace(".elf", "") + ".bin"
 # Convert the ELF executable to a binary format
-subprocess.run(["aarch64-none-elf-objcopy", "-O", "binary", elf_executable, bin_executable_location], check=True)
```
arihant2math, line 26: Any reason for this change? It breaks CI as well as my build.

src/kernel/rand.rs
```diff
@@ -1,39 +1,227 @@
...
+                WakeupType::All
+            });
+        }
+    }
+
+    /// Add entropy into this pool based upon the uptime of the system.
+    ///
+    /// SAFETY: This method should *not* be called for predictable or periodic
+    /// events.
+    pub unsafe fn add_temporal_entropy(&self) {
```
arihant2math: Even though this method is dangerous, I'm not sure that it should be marked as unsafe, since it doesn't affect the overall program's safety in any way.

src/drivers/chrdev/random.rs
```diff
@@ -34,7 +35,10 @@ impl FileOps for RandomFileOps {
     }
 
     async fn readat(&mut self, buf: UA, count: usize, _offset: u64) -> Result<usize> {
-        sys_getrandom(buf.cast(), count as _, 0).await
+        let mut kbuf = vec![0u8; count];
+        fill_random_bytes(&mut kbuf).await;
```
arihant2math, line 41: Something to think about is that `/dev/random` tries to wait until entropy is high enough, but `/dev/urandom` behaves as this is implemented.

scripts/qemu_runner.py
hexagonal-sun: No, sorry, that snook in!

src/kernel/rand.rs
hexagonal-sun: You're correct, it isn't UB to call this funciton incorrectly, just a security issue which isn't what `unsafe` is for.

src/drivers/chrdev/random.rs
hexagonal-sun: You're right, `/dev/random` never blocks... this is more like the way `/dev/random` behaves.   I'll add a `TODO`.

PR #253, opened 2026-03-07
Fix linux incompatibilities in sockets API

arihant2math:
- Add filtration to `usertest` to only run specific tests (Usage: `usertest <substr>`)
- rename `socket` dir to `net`
- Add `sys_accept4` and ensure `sys_accept` actually sets socket addr
- Allow `MSG_NOSIGNAL` for `sys_sendto` because we don't send a signal anyhow.
- Implement `sys_recvfrom`
- Update syscall table

This gets the `test_rust_unix_socket` test working, which is a non-contrived test of the capabilities of UDS stream sockets.


hexagonal-sun (approved):
LGTM!

Issue #252, opened 2026-03-07
Pasting more than 32 bytes of text into the UART causes an indefinite hang


arihant2math:
For example, pasting `strace -f /bin/usertest test_unix_socket`.

arihant2math:
Seems like it can't handle more than 32 characters?

PR #255, opened 2026-03-08
Proper draining of uart to fix hang

arihant2math:
Support up to 4096 char copies at a time (instead of the current 32).

Fixes: #252


hexagonal-sun (approved):
LGTM!

PR #258, opened 2026-03-11
Update ext4plus to 0.1.0-alpha.5

arihant2math:
Full changelog: https://github.com/arihant2math/ext4plus/blob/main/CHANGELOG.md

Mostly cleanup, but fixes a few bugs that caused fsck issues.

- Makes the `Dir` API more like the `File` API
- Expose `Inode::blocks` and `Inode::set_blocks`.
- Expose `Inode::crtime` and `Inode::set_crtime` as well as `Metadata.crtime` (`Option<Duration>`, due to lack of availability on 32-bit filesystems). (unused for now)
- Remove all restrictions on block map writing, allowing writing to block maps even with indirect blocks.
- Support extra 2 bits for seconds in timestamp to avoid Y2038 problem as well as nanosecond precision.
- Fix unused_inodes_count not being updated when allocated inodes, leading to inodes possibly being overwritten by other drivers.
- Update `i_blocks` on writes.
- Set `extra_size` correctly when creating an inode
- Handle non 128/256-byte inode sizes better

In the works: xattr support, more fsck fixes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,9 +256,9 @@ dependencies = [
 
 [[package]]
 name = "ext4plus"
-version = "0.1.0-alpha.4"
+version = "0.1.0-alpha.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7d4c717f942d8f2d0ba620e17afac4fd8c2c9ad315ca205d652bcd4c0503ed1b"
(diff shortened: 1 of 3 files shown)


arihant2math:
There were a few issues with the 4th alpha, so I released a 5th alpha with bugfixes.

PR #259, opened 2026-03-12
Implement `/proc/<pid>/exe`

arihant2math:
Used by zig.

### DIFF
--- a/src/drivers/fs/proc/task/task_file.rs
+++ b/src/drivers/fs/proc/task/task_file.rs
@@ -18,6 +18,7 @@ pub enum TaskFileType {
     State,
     Stat,
     Maps,
+    Exe,
 }
 
 impl TryFrom<&str> for TaskFileType {
@@ -32,6 +33,7 @@ impl TryFrom<&str> for TaskFileType {
             "cwd" => Ok(TaskFileType::Cwd),
             "root" => Ok(TaskFileType::Root),
             "maps" => Ok(TaskFileType::Maps),
+            "exe" => Ok(TaskFileType::Exe),
             _ => Err(()),
         }
     }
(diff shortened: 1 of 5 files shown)


src/drivers/fs/proc/task/task_file.rs
```diff
@@ -243,6 +253,28 @@ Threads:\t{tasks}\n",
...
+                .map(|(desc, _)| *desc);
+            drop(task_list);
+            let task_details = if let Some(desc) = id {
+                find_task_by_descriptor(&desc)
+            } else {
+                None
+            };
+            return if let Some(task) = task_details {
+                if let Some(exe) = task.process.executable.lock_save_irq().clone() {
+                    Ok(exe.as_str().to_string().into())
```
hexagonal-sun, line 271: I think linux returns the `exe` as a symlink rather than a string.   Could we model that behavior?

src/process/thread_group/builder.rs
```diff
@@ -90,6 +90,7 @@ impl ThreadGroupBuilder {
             next_tid: AtomicU32::new(1),
             state: SpinLock::new(ProcessState::Running),
             tasks: SpinLock::new(BTreeMap::new()),
+            executable: SpinLock::new(None),
```
hexagonal-sun, line 93: I don't see where during a `clone()` the new child process would copy the executable from it's parent.  I think during a `fork()` the executable would be set to `None`.

src/drivers/fs/proc/task/task_file.rs
arihant2math: This is in the `readlink` method.

src/process/exec.rs
```diff
@@ -212,6 +213,7 @@ async fn exec_elf(
     let mut fd_table = current_task().fd_table.lock_save_irq().clone();
     fd_table.close_cloexec_entries().await;
     *current_task().fd_table.lock_save_irq() = fd_table;
+    *current_task().process.executable.lock_save_irq() = Some(path.to_owned());
```
arihant2math, line 216: I do that here

hexagonal-sun (approved):
LGTM

PR #260, opened 2026-03-13
sched refactor

hexagonal-sun:
- **sched: introduce Work as the unified scheduleable unit**
  Refactor the scheduler so all scheduleable work is wrapped in Arc<Work>,
  replacing the previous per-CPU wait_q design where sleeping tasks were
  bound to a specific CPU. Wakers now hold direct Arc<Work> references and
  can re-enqueue tasks on any CPU upon wakeup.
  
  Key changes:
  
  - Add Work struct wrapping OwnedTask with an AtomicTaskState and
    scheduler metadata (SchedulerData), replacing the old SchedulableTask.
    Remove Task::state (Arc<SpinLock<TaskState>>). Work::state is now the
    single source of truth for task state.
  
  - Rewrite the run queue using BinaryHeap-based eligible/ineligible split
    (EEVDF) with a dedicated VClock, replacing the BTreeMap linear scan.
    Extract vclock into its own module.
  
  - Rewrite wakers to hold Arc<Work> directly instead of looking up tasks
    by TaskDescriptor from TASK_LIST.
  
  - Replace lock-based sleep transitions in uspc_ret with atomic CAS
    (try_sleep_current) that correctly detects concurrent Woken state.
  
  - Simplify least-tasked-CPU metric to use only run-queue weight, since
    sleeping tasks are no longer bound to any CPU.
  
  - Add current_work() accessor.
  

- **sched: runqueue: derfer task drops and drop Finished tasks**
  Fix two issues:
  
  1. When a task is dropped from the runqueue, it will trigger the
     destructors for the task to run. This may well call wakers to wake up
     parent processes, other ends of pipes, etc. If we do that while
     `SCHED_STATE` is still borrowed, this causes a double-borrow panic. Fix
     this by deferring all drops until after we have unlocked `SCHED_STATE`.
  
  2. Tasks inside the runqueue which are yet to be scheduled which become
     finished will be returned by `find_next_task` and the state will be set
     to `TaskState::Running` overwriting the fact that this task had
     `Finish`ed. We'd then queue this task forever. Filter finished tasks in
     `find_next_task` and add them to the defered drop list.
  

- **sched: runqueue: fix tick and runqueue insertion bug**
  Ensure that `tick()` is called on the current task, and allow the task
  account decide whether we should try to switch to another task.
  
  Also, ensure accounting is updated for freshly inserted tasks into the
  runqueue.

### DIFF
--- a/src/sched/runqueue/mod.rs
+++ b/src/sched/runqueue/mod.rs
@@ -96,15 +96,21 @@ impl RunQueue {
         self.v_clock.advance(now, self.weight());
 
         let mut prev_task = ptr::null();
+        let mut next_task = None;
         let mut deferred_drops: Vec<RunnableTask> = Vec::new();
 
-        if let Some(cur_task) = self.running_task.take() {
+        if let Some(mut cur_task) = self.running_task.take() {
+            prev_task = Arc::as_ptr(&cur_task.task);
             let state = cur_task.task.state.load(Ordering::Acquire);
             match state {
                 TaskState::Running | TaskState::Woken => {
-                    // requeue for the next time slice.
-                    prev_task = Arc::as_ptr(&cur_task.task);
-                    self.enqueue(cur_task);
+                    if cur_task.tick(now) {
+                        // Deadline exceeded — requeue for the next time slice.
+                        self.enqueue(cur_task);
+                    } else {
+                        // Still has budget — keep running.
+                        next_task = Some(cur_task);
+                    }
                 }
                 TaskState::PendingSleep | TaskState::PendingStop => {
                     // Task wants to deactivate. Drop the RunnableTask now to
@@ -126,7 +132,8 @@ impl RunQueue {
             }
         }
 
-        if let Some(mut next_task) = self.find_next_task(&mut deferred_drops) {
+        if let Some(mut next_task) = next_task.or_else(|| self.find_next_task(&mut deferred_drops))
+        {
             next_task.about_to_execute(now);
 
             if Arc::as_ptr(&next_task.task) != prev_task {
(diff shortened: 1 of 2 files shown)


src/sched/mod.rs
```diff
@@ -128,12 +131,10 @@ pub fn spawn_kernel_work(fut: impl Future<Output = ()> + 'static + Send) {
 
 /// Global atomic storing info about the least-tasked CPU.
 /// First 16 bits: CPU ID
-/// Next 24 bits: Weight
-/// Next 24 bits: Number of waiting tasks
+/// Remaining 48 bits: Run-queue weight
```
arihant2math, line 134: The reason I included the number of waiting tasks was to prevent a single CPU from taking on all tasks because they were bursty or I/O bound. Ideally we'd rather schedule onto a CPU with less waiting tasks, all else equal. Maybe we should consider total weight (run queue + waiting queue).

Not something that needs to change in the PR.

arihant2math: Also probably worth removing this atomic. I think linux does proper searching. However, for this we'd need a concept of "shared scheduler state".

src/sched/sched_task/state.rs
```diff
@@ -0,0 +1,240 @@
...
+    PreventedSleep,
+    /// No action needed (already Woken/Finished, or CAS lost to another CPU).
+    None,
+}
+
+/// State machine wrapper around `AtomicTaskState`.
+///
+/// Only exposes named transition methods to enforce state transition logic.
+///
+/// ```text
```
arihant2math, line 55: nice diagram :)

arihant2math (commented):
LGTM

src/sched/sched_task/state.rs
hexagonal-sun: Thanks.. if you use Emacs have a look at picture mode!

PR #261, opened 2026-03-14
Retain CPU affinity when rescheduling

arihant2math:
Prevent cache invalidation. Task stealing should be easy to implement once an algorithm is in place.

Also the weird static logic has been replaced with a shared per cpu struct for simplicity (also because in the future it'd be good to prefer CPUs that share the same L3 cache/numa node/etc.).

### DIFF
--- a/src/sched/mod.rs
+++ b/src/sched/mod.rs
@@ -1,4 +1,5 @@
-use crate::drivers::timer::{Instant, now};
+use crate::arch::{Arch, ArchImpl};
+use crate::drivers::timer::now;
 #[cfg(feature = "smp")]
 use crate::interrupts::cpu_messenger::{Message, message_cpu};
 use crate::kernel::cpu_id::CpuId;
@@ -70,6 +71,10 @@ per_cpu_private! {
     static SCHED_STATE: SchedState = SchedState::new;
 }
 
+per_cpu_shared! {
+    pub static SHARED_SCHED_STATE: SharedSchedState = SharedSchedState::new;
+}
+
 /// Default time-slice assigned to runnable tasks.
 const DEFAULT_TIME_SLICE: Duration = Duration::from_millis(4);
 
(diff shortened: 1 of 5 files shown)


src/process/mod.rs
```diff
@@ -159,6 +162,8 @@ pub struct Task {
     pub utime: AtomicUsize,
     pub stime: AtomicUsize,
     pub last_account: AtomicUsize,
+    pub last_cpu: AtomicUsize,
+    pub cpu_mask: SpinLock<[u64; CPU_MASK_SIZE]>,
```
hexagonal-sun: I'd put these members in `struct Work`, since that's where scheduling info is located.

src/sched/mod.rs
```diff
@@ -129,18 +134,22 @@ pub fn spawn_kernel_work(fut: impl Future<Output = ()> + 'static + Send) {
...
-const WEIGHT_SHIFT: u32 = 16;
-
 #[cfg(feature = "smp")]
 fn get_best_cpu() -> CpuId {
-    // Get the CPU with the least number of tasks.
-    let least_tasked_cpu_info = LEAST_TASKED_CPU_INFO.load(Ordering::Acquire);
-    CpuId::from_value((least_tasked_cpu_info & 0xffff) as usize)
+    let r = 0..ArchImpl::cpu_count();
+    r.min_by(|&x, &y| {
+        // TODO: Find a way to calculate already assigned affinities and account for that
```
hexagonal-sun, line 141: Yeah, this is complicated since runqueues no longer own the tasks that they are waiting on .   I suspect we'll end up using some kind of atomic count.

src/sched/mod.rs
```diff
@@ -198,27 +212,15 @@ impl SchedState {
...
-        }
+        let current_task = self.run_q.current().task.descriptor().tid();
+        SHARED_SCHED_STATE
+            .get()
+            .current_task_id
+            .store(current_task.value(), Ordering::Relaxed);
+        SHARED_SCHED_STATE
+            .get()
+            .total_runq_weight
+            .store(weight, Ordering::Relaxed);
```
hexagonal-sun: This function is called at the beginning of `do_schedule()`.  Since it's invoked *before* the schedule logic is called won't that mean that this info doesn't reflect the current state?

src/sched/mod.rs
```diff
@@ -245,6 +247,20 @@ impl SchedState {
     }
 }
 
+pub struct SharedSchedState {
+    pub current_task_id: AtomicU32,
```
hexagonal-sun: Where is this used?  I don't see why this would be useful.  We can access the current task `Arc<Work>` on the current CPU with `current_work()`.

src/process/mod.rs
hexagonal-sun: In fact, perhaps they could live in `struct SchedInfo`.  That way if a syscall wants to adjust the CPU affinity, it could call something like `current_running_task()` to get a mutable ref to the currently scheduled task (which contains a `SchedInfo`).

PR #263, opened 2026-03-15
Implement `procs_running` and `procs_blocked` in `/proc/stat`

arihant2math:


src/drivers/fs/proc/stat.rs
```diff
@@ -82,6 +84,26 @@ impl SimpleFile for ProcStatInode {
...
+                procs_running += 1;
+            } else if matches!(
+                state,
+                TaskState::Sleeping
+                    | TaskState::Stopped
+                    | TaskState::PendingSleep
+                    | TaskState::PendingStop
+            ) {
+                procs_blocked += 1;
+            }
```
hexagonal-sun: Maybe using `match state` would be cleaner here?

PR #274, opened 2026-03-25
Implement itimer syscalls with generic timer handling framework

arihant2math:
Replaces #267 with a generic callback framework to handle arbitrary timers.

> Interval timers are an POSIX feature and come in three forms (real, virtual, prof). This PR only implements itimers on the wall clock.
> 
> I've moved all the clock syscalls to `src/clock/syscalls/` 
> 
> Known issues/followups/etc.:
> - We don't move tasks between CPUs, but if we did, removing an itimer might fail because removing from non-local wakeup queues isn't supported.
> - No support for virtual/prof, they likely can be implemented by doing a scheduler check before starting a task, however that requires some infrastructure.
> - Other timers depend on the current date, which can be adjusted and is different from wall time, so we need a separate wakeup queue for that.
> - Linux helps prevent interrupt spam by preventing interrupts from being scheduling too early in the future (i.e. interrupts must be scheduled atleast X nanoseconds in the future) and by grouping interrupts that are within ~50ns of each other, we probably should do this too.
> 
> Testing: usertest


src/clock/syscalls/itimer.rs
```diff
@@ -0,0 +1,196 @@
...
+                timers.real = None;
+                None
+            }
+        }
+        _ => unimplemented!(),
+    }
+}
+
+async fn getitimer(current_task: &Task, which: ITimerType) -> libkernel::error::Result<ITimerVal> {
+    let now = match which {
```
hexagonal-sun, line 99: Rather than this being a duration, wouldn't it be cleaner if this was an instant?  So maybe rather than using `uptime()` below, maybe use `now()`?

src/clock/syscalls/itimer.rs
```diff
@@ -0,0 +1,196 @@
...
+                        current_task.tid(),
+                        make_timer_id(TimerNamespace::ITimer, ITimerType::Real as u32),
+                    );
+            }
+            if let Some(next) = next {
+                timers.real = Some(ITimer { interval, next });
+                crate::drivers::timer::SYS_TIMER
+                    .get()
+                    .unwrap()
+                    .schedule_timer(
```
hexagonal-sun, line 188: Should we also remove scheduled timers on task exit?   in `exit.rs`, just iterate the namespace and call `remove_scheduled_timer()` for any that are set.

src/process/clone.rs
```diff
@@ -180,6 +180,7 @@ pub async fn sys_clone(
                 fd_table: files,
                 cwd,
                 root,
+                i_timers: SpinLock::new(*current_task.i_timers.lock_save_irq()),
```
hexagonal-sun: Is this correct?  According to the man page for `timer_create()`:

```
Timers are not inherited by the child of a [fork(2)](https://man7.org/linux/man-pages/man2/fork.2.html), and are
       disarmed and deleted during an [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html).
```

hexagonal-sun (changes requested):

arihant2math:
Good points about edge-cases, adjusted the code so that these are handled as needed.

hexagonal-sun (approved):
LGTM!

PR #279, opened 2026-04-05
Implement setuid, setgid, setreuid, setregid, setresuid, setresgid

arihant2math:
If there is a neat way to clean the repetition up here, that'd be great.


hexagonal-sun (approved):
LGTM, as you say.. if there's a way to improve the amount of repetition, that would be good.  Can't think of anything off-hand, though.

PR #282, opened 2026-04-14
Implement signalfd4 and epoll for systemd

arihant2math:
Both are used together for the purposes of systemd, so implementing them together made for easy testing.

Testing: usertest and systemd

### DIFF
--- a/src/fs/pipe.rs
+++ b/src/fs/pipe.rs
@@ -9,6 +9,8 @@ use crate::{
     sched::{current_work, syscall_ctx::ProcessCtx},
     sync::CondVar,
 };
+use core::pin::Pin;
+
 use alloc::{boxed::Box, sync::Arc};
 use async_trait::async_trait;
 use core::any::Any;
@@ -120,6 +122,29 @@ impl PipeReader {
 
 #[async_trait]
 impl FileOps for PipeReader {
+    fn poll_read_ready(&self) -> Pin<Box<dyn Future<Output = Result<()>> + 'static + Send>> {
+        let inner = self.inner.clone();
+        Box::pin(async move {
+            let mut read_fut = Box::pin(inner.buf.read_ready());
+            let mut gone_cond = Box::pin(
+                inner
+                    .other_side_gone
+                    .wait_until(|gone| if *gone { Some(()) } else { None }),
+            );
+
+            future::poll_fn(move |cx| {
+                if read_fut.as_mut().poll(cx).is_ready() {
+                    Poll::Ready(Ok(()))
+                } else if gone_cond.as_mut().poll(cx).is_ready() {
+                    Poll::Ready(Ok(()))
+                } else {
+                    Poll::Pending
+                }
+            })
+            .await
+        })
+    }
+
     async fn read(&mut self, _ctx: &mut FileCtx, u_buf: UA, count: usize) -> Result<usize> {
         self.readat(u_buf, count, 0).await
     }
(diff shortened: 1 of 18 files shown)


src/fs/pipe.rs
```diff
@@ -120,6 +122,27 @@ impl PipeReader {
...
+        let inner = self.inner.clone();
+        Box::pin(async move {
+            let mut read_fut = Box::pin(inner.buf.read_ready());
+            let mut gone_cond = Box::pin(
+                inner
+                    .other_side_gone
+                    .wait_until(|gone| if *gone { Some(()) } else { None }),
+            );
+
+            future::poll_fn(move |cx| {
```
hexagonal-sun: I wonder whether it's worth adding a `select!` macro to libkernel.  I expect we'll be doing a lot of polling of two futures.

src/process/thread_group/signal/signalfd.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        loop {
+            if let Some(sig) = self.take_pending_signal() {
+                let info = SignalfdSiginfo {
+                    ssi_signo: sig.user_id() as u32,
+                    ..Default::default()
+                };
+
+                let sig_tua = TUA::<SignalfdSiginfo>::from_value(ptr.value());
+                copy_to_user(sig_tua, info).await?;
```
hexagonal-sun, line 128: You might be able to do `buf.cast()` here.

src/process/thread_group/signal/signalfd.rs
```diff
@@ -0,0 +1,255 @@
...
+
+                if bytes_read + siginfo_size > count {
+                    break;
+                }
+            } else if bytes_read > 0 {
+                break;
+            } else if nonblock {
+                return Err(KernelError::TryAgain);
+            } else {
+                crate::drivers::timer::sleep(core::time::Duration::from_millis(10)).await;
```
hexagonal-sun: Rather than doing this and 'polling' in a loop.  Maybe use a CondVar over the SigSet, then you get proper wake ups when something modifies the sigset.

src/process/thread_group/signal/signalfd.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    fn poll_read_ready(&self) -> Pin<Box<dyn Future<Output = Result<()>> + 'static + Send>> {
+        let mask = self.mask;
+        Box::pin(async move {
+            let signalfd = SignalFd { mask };
+            loop {
+                if signalfd.has_pending_signal() {
+                    return Ok(());
+                }
+                crate::drivers::timer::sleep(core::time::Duration::from_millis(10)).await;
```
hexagonal-sun: Likewise.

src/process/thread_group/signal/signalfd.rs
hexagonal-sun: Ah wait, I see.  It looks at the current tasks pending signals and takes those.   Now I don't implement any kind of notification mechanism for signal delivery since it's actioned by the task when it is next scheduled.  Hmm.    There's only one place where we raise a signal `raise_task_signal`.  One option is to wrap the Task sigset in a condvar?  That you can get to receieve on-demand notifications on it. Although, it's an atomic.... Hmm, another option is to not wrap the sig set, but maybe have another field in the Task, something like signal_notifier: WakerSet<()>.    Then you get dynamic notifications.

arihant2math:
Fixed the busy-waiting issue.

PR #283, opened 2026-04-16
libkernel/x86 64 paging

hexagonal-sun:
- **libkernel: arm64: memory: add a MemoryType AT to PaMapper**
  Allow implementors of the `PaMapper` trait to choose a type which
  represents a memory type when mapping a PA region.
  

- **libkenrel: memory: move paging-related modules to paging module**
  Move PtePermissions to paging module which is feature-gated behind
  `paging`.
  
  Also move all AddressSpace related functionality behind the `proc_vm`
  feature gate.
  

- **libkernel: move paging-related arch-agnostic traits out of arch mod**
  Any traits which are architecture agnostic should be available to be
  implemented by other architectures. Move them out from the arm64 module
  into a non-architecure specific location.
  

- **libkernel: x86_64: pg_descriptors: new**
  Add an implementation of x86_64 page descriptors (PML4E, PDPE, PDE,
  PTE). Also add corresponding unit tests to ensure correct functionality.
  

- **libkernel: memory: paging: make PageTableEntry contain MAP_SHIFT**
  We currently define the MAP_SHIFT for a page-table level in two places,
  `TableMapper` and `PgTable`. Store only one shift trait constant value
  in the `PageTableEntry` trait.
  

- **libkernel: paging: move test harness code out of arm64**
  The test harness code is useful across architectures. Therefore, move it
  out of the arm64 directory so it can be used by other architectures
  implementing paging unit-tests.
  

- **libkernel: x86_64: pg_tables: new**
  Implement page table module for x86_64.
  

- **libkernel: x86_64: fix impl of bit 7**
  Bit 7 on various descriptors in x86_64 can take on varying values
  depending upon the descriptor type. Refelect that in the implementation
  and fix a bug whereby the PAT bit would have been set when creating a
  PTE.
  

- **libkernel: paging: move NullTlbInvalidator into shared code**
  Move the NullTlbInvalidator implementation into shared code since it's
  architecturally agnostic.
  

- **libkernel: memory: move RecursiveWalker to non-arch module**
  Since the architectural details of descriptor implementation is hidden
  behind traits, the code for walking page tables is arch-agnostic.
  Therefore, promote it to shared code for use by other architectures.
  

- **libkernel: paging: move NextLevel to TableMapper trait**
  Move the NextLevel AT into the TableMapper (descriptor-level) trait.
  This allows us to provide a blanket implementation of TableMapperTable
  for all tables where, Table::Descriptor : TableMapper.
  

- **libkernel: paging: x86_64: implement pg_walk**
  Implement the pg_walk functionality for x86_64.
  

- **libkernel: paging: x86_64: add canonical address tests**
  Ensure that working with canonical addresses (sign extension of bit 47)
  works correctly.
  

- **libkernel: x86_64: add pg_teardown module**
  Add a module for tearing down x86_64 userspace page table hierarchies.


libkernel/src/arch/x86_64/memory/pg_tear_down.rs
```diff
@@ -0,0 +1,287 @@
...
+use crate::memory::{
+    address::{PA, TPA},
+    paging::{
+        PaMapper, PageTableMapper, PgTable, PgTableArray, tear_down::RecursiveTeardownWalker,
+        walk::WalkContext,
+    },
+};
+
+// Implementation for PTable (Leaf Table)
+impl RecursiveTeardownWalker for PTable {
```
arihant2math, line 17: The teardown walker only follows `next_table_address() `and the x86 leaf implementation only frees payloads from `PTable`. A userspace address space containing a 2 MiB mapping will free the page-table pages/root table while leaking the mapped physical frames. Teardown needs to account for valid block descriptors at non-leaf levels I think.

libkernel/src/arch/x86_64/memory/pg_tables.rs
```diff
@@ -0,0 +1,1185 @@
...
+/// 2. The `allocator` and `mapper` provided in the `ctx` must be correctly
+///    implemented and functional.
+/// 3. Concurrency: The caller must ensure that no other CPU core is
+///    concurrently accessing or modifying the page tables being manipulated.
+///    This typically requires holding a spinlock that guards the entire address
+///    space.
+/// 4. Memory Lifecycle: The caller is responsible for managing the lifecycle of
+///    the physical memory described by `attrs.phys`. This function only creates
+///    a *mapping* to the memory; it does not take ownership of it.
+pub fn map_range<PA, PM>(
```
arihant2math, line 182: Should there be a check to make sure the address is canonical? From what it seems the mapper will install a mapping for a different canonical address instead of
 rejecting the input in the arch agnostic code (libkernel/src/memory/paging/mod.rs:119).

arihant2math (commented):
LGTM minus these two issues.

libkernel/src/arch/x86_64/memory/pg_tables.rs
hexagonal-sun: Good point.  I should probably reject non-canonical VAs.   I'll add that check.

libkernel/src/arch/x86_64/memory/pg_tear_down.rs
hexagonal-sun: Agreed.  I've pushed a change to support tearing down block mappings.

PR #286, opened 2026-04-21
libkernel: pg_walk: use saturating arithmetic to avoid overflow at last page-table entry

hexagonal-sun:
The generic RecursiveWalker computed the coverage region for each page-table entry as:

    VirtMemoryRegion::new(entry_va, table_coverage).intersection(region)

For the last entry at the PML4 level (PML4[511]) in the kernel high-half, entry_va + table_coverage wraps past usize::MAX, causing an arithmetic overflow panic inside end_address().

Fix this in walk.rs by computing the intersection bounds with saturating_add/min/max directly.

Add regression tests for all levels, ensuring last-element walking is correct.

### DIFF
--- a/libkernel/src/arch/x86_64/memory/pg_walk.rs
+++ b/libkernel/src/arch/x86_64/memory/pg_walk.rs
@@ -391,6 +391,101 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


arihant2math (approved):
LGTM

PR #292, opened 2026-05-06
libkernel improvements

hexagonal-sun:
- **libkernel: CpuOps: make interrupt flags a associated type**
  Rather than hard-coding the interrupt flags as a `usize`. Allow each CPU
  architecture to define it's own interrupt flags type via an associated
  type.
  

- **libkernel: pg_offset: decouple from `VirtualMemory` trait**
  Since the only requirement for a `PageOffsetTranslator` is a constant
  offset, pass that in as a generic parameter, allowing it to be decoupled
  from the `VirtualMemory` trait.
  

- **libkernel: move `SmallocPageAlloc` into libkernel**
  The `SmallocPageAlloc` type is generic and architecture-independent, so
  it belongs in libkernel rather than the kernel's arm64-specific code. It
  is placed in memory::paging::smalloc_page_allocator and gated by the
  existing paging feature.
  
  The kernel's copy is replaced with a thin re-export so all call sites
  are unaffected.
  

- **libkernel: pg_tear_down: add more flexibility**
  The current tear down implementation assumes that all data mappings &
  intermediate tables need to be free'd. That might not provide enough
  flexibility for all tear-down scenarios.
  
  Provide a more flexible API, passing current walk state into a control
  function which makes a descision regarding tear-down and recursion.
  

- **libkernel: smalloc: add claim_region for downstream allocators**
  Introduce Smalloc::claim_region, which removes a sub-region from the
  free memory pool and transfers ownership to another allocator. The
  region is punched out of the memory list rather than being added to the
  reservation list.
  

- **libkernel: phys: support managing a sub-region of memory**
  Add `FrameAllocator::init_from_region`, which constructs a buddy frame
  allocator over an explicit PhysMemoryRegion rather than the entire
  memory span known to smalloc. This is a building block for managing
  multiple disjoint memory pools (e.g. per-NUMA-node allocators).
  

- **libkernel: memory: phys: expose MAX_ORDER const.**
  Expose the physical memory allocator MAX_ORDER const.


arihant2math (approved):
LGTM

PR #295, opened 2026-05-11
Support bringing up block device as rootfs

arihant2math:
- Adds the concept of disks that can be mounted and unmounted (exposed in /dev).
- Adds support for virtio disks drivers. This allows for bringing up large filesystems.
- To allow for pivot_root to function this also required the implementation of `pivot_root` and `unmount`.

Running the following commands will mount the disk and switch to ubuntu:
```sh
mkdir /new-root
mount -t ext4 /dev/vda /new-root

# Create a mount point for the old root inside the new root
mkdir -p /new-root/old-root

cd /new-root

# Mount devfs in the new root so that console is available
mkdir -p dev
mount -t devfs devfs dev

pivot_root . old-root

# change to ubuntu
exec chroot . /bin/bash <dev/console >dev/console 2>&1
```

Future work:
- intramfs?
- Allow for directly loading from a disk.
- Ubuntu's `/bin/ls` has some weird crashing issue involving segfaults and `/bin/bash` has issues with stack smashing.


src/arch/arm64/exceptions/syscall.rs
```diff
@@ -616,6 +620,7 @@ pub async fn handle_syscall(mut ctx: ProcessCtx) {
         0x9b => sys_getpgid(&ctx, arg1 as _),
         0x9c => sys_getsid(&ctx).await,
         0x9d => sys_setsid(&ctx).await,
+        0x9e => sys_getgroups(&ctx, arg1 as _, TUA::from_value(arg2 as _)).map_err(|e| match e {}),
```
hexagonal-sun, line 623: This this one might be missing from the markdown file?

src/drivers/virtio_hal.rs
```diff
@@ -16,6 +25,87 @@ impl VirtioHal {
         let rounded = pages.next_power_of_two();
         rounded.ilog2() as u8
     }
+
+    fn translated_phys_addr(vaddr: VA) -> Option<PhysAddr> {
+        ArchImpl::kern_address_space()
+            .lock_save_irq()
+            .translate(vaddr)
```
hexagonal-sun: FYI, this might want to use the new `translate()` API that I've created in libkernel.   It supports THP so block mappings work too!

src/drivers/virtio_hal.rs
```diff
@@ -16,6 +25,87 @@ impl VirtioHal {
...
+            BufferDirection::DriverToDevice | BufferDirection::Both
+        ) {
+            Self::bounce_copy_in(paddr, buffer);
+        }
+
+        BOUNCED_SHARES
+            .lock_save_irq()
+            .push(BouncedShare { paddr, pages });
+
+        // trace!(
```
hexagonal-sun: Debug?

src/drivers/virtio_hal.rs
```diff
@@ -40,12 +130,12 @@ unsafe impl Hal for VirtioHal {
             core::ptr::write_bytes(vaddr.as_ptr(), 0, pages * PAGE_SIZE);
         }
 
-        trace!("alloc DMA: paddr={paddr:#x}, pages={pages}, order={order}");
+        // trace!("alloc DMA: paddr={paddr:#x}, pages={pages}, order={order}");
```
hexagonal-sun: Debug?

hexagonal-sun: There's other commented debug stmts too.

src/fs/syscalls/pivot_root.rs
```diff
@@ -0,0 +1,138 @@
...
+        if root.0.id() == old_root_inode.id() {
+            *root = (new_root_inode.clone(), "/".into());
+        }
+        drop(root);
+
+        let mut cwd = task.cwd.lock_save_irq();
+        if cwd.0.id() == old_root_inode.id() {
+            *cwd = (new_root_inode.clone(), "/".into());
+        }
+    }
```
hexagonal-sun: Do we also need to worry about the file tables for tasks too?   How does that work? I presume we'd need to iterate through all open FDs.

Thinking about it though, this sounds very complicated.  Does Linux just keep those FDs open but any *new* FDs are from the new root?

src/fs/syscalls/umount.rs
```diff
@@ -0,0 +1,62 @@
...
+    let cwd = task.cwd.lock_save_irq().0.clone();
+    let target_path = Path::new(target);
+    let target = if flags.contains(UmountFlags::UMOUNT_NOFOLLOW) {
+        VFS.resolve_path_nofollow(target_path, cwd, &task).await?
+    } else {
+        VFS.resolve_path(target_path, cwd, &task).await?
+    };
+
+    VFS.umount(target, flags.contains(UmountFlags::MNT_DETACH))
+        .await?;
```
hexagonal-sun, line 59: Do we have to worry about any open files at the location being umounted?

hexagonal-sun (changes requested):

src/fs/syscalls/pivot_root.rs
arihant2math: I think linux just ignores any existing FDs, cwds, or roots. I've removed all the logic relating to that.

src/fs/syscalls/umount.rs
arihant2math: `VFS::umount` handles this.

arihant2math:
@hexagonal-sun mind taking another look when you get the chance?

PR #299, opened 2026-05-12
systemd part 7: implement shared mmap maps

arihant2math:
Cherry-picked out of the stack; implements shared maps for mmap.

### DIFF
--- a/src/memory/mmap.rs
+++ b/src/memory/mmap.rs
@@ -1,12 +1,15 @@
 use core::sync::atomic::{AtomicUsize, Ordering};
 
-use crate::{process::fd_table::Fd, sched::syscall_ctx::ProcessCtx};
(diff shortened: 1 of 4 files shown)


src/memory/mmap.rs
```diff
@@ -128,14 +177,23 @@ pub async fn sys_mmap(
...
         requested_len,
         permissions,
         kind,
         name,
+        is_shared,
     )?;
 
+    if is_shared
+        && let Err(e) =
+            populate_shared_anon_mapping(ctx, new_mapping_addr, requested_len, permissions).await
```
hexagonal-sun, line 191: Any reason we don't demand-page here?    I'm not too sure about the semantics of a shared mapping, is it preallocated up-front?

PR #312, opened 2026-06-10
Check for LLVM and Rust objcopy binaries before throwing

sixseventysix:
The `qemu_runner.py` file hardcoded `aarch64-none-elf-objcopy` for the `objcopy` binary, which is distributed as part of the Arm GNU toolchain, and isn't available by default on macOS. LLVM ships with its own objcopy binary that has support for multiple targets and is described as a "drop-in replacement for GNU's objcopy" (see: https://llvm.org/docs/CommandGuide/llvm-objcopy.html).
The `qemu_runner.py` now checks if the system has `llvm-objcopy` and `rust-objcopy` (cargo wrapper over `llvm-objcopy`) and invokes whichever one it finds. If it doesn't find any, it throws an error instructing the programmer to install any of the above objcopy binaries.

### DIFF
--- a/scripts/qemu_runner.py
+++ b/scripts/qemu_runner.py
@@ -1,7 +1,14 @@
 #!/usr/bin/env python3
 
 import argparse
+import shutil
 import subprocess
+
+def find_objcopy():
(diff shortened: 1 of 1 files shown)


arihant2math (approved):
lgtm

PR #313, opened 2026-06-30
futex2

sixseventysix:


arihant2math:
I've fixed CI on master, do you mind rebasing?

sixseventysix:
> I've fixed CI on master, do you mind rebasing?

just a reminder in case you missed it. I rebased my branch on master and CI is now green

usertest/src/futex2.rs
```diff
@@ -0,0 +1,736 @@
...
+    }
+}
+
+register_test!(test_futex2_wake_no_waiters);
+
+fn test_futex2_requeue_to_self() {
+    // Requeueing a futex onto itself (uaddr1 == uaddr2) is rejected with
+    // EINVAL on Linux. moss currently short-circuits this to a plain wake and
+    // returns success, silently dropping nr_requeue -- this test encodes the
+    // Linux contract and is expected to fail until that is fixed.
```
hexagonal-sun: Think this comment may be outdated?  The test asserts `EINVAL` below, Unless this test is actually still failing (but I'm assuming not since CI is green)?

src/clock/mod.rs
```diff
@@ -3,6 +3,86 @@ pub mod syscalls;
...
+/// wait is in progress.
+#[derive(Clone, Copy)]
+pub enum Deadline {
+    /// Absolute instant on the monotonic clock (`CLOCK_MONOTONIC`).
+    Monotonic(Duration),
+    /// Absolute instant on the realtime clock (`CLOCK_REALTIME`).
+    Realtime(Duration),
+}
+
+impl Deadline {
```
arihant2math, line 28: I think a good follow up PR would be to use `Deadline` in more places for more support for sleeping on realtime clock

src/clock/mod.rs
```diff
@@ -3,6 +3,86 @@ pub mod syscalls;
...
+    ///
+    /// The monotonic clock advances uniformly, so a single relative sleep is
+    /// exact. The realtime clock can be stepped by `clock_settime`, so a
+    /// realtime wait races the timer against a clock-was-set notification: on
+    /// either it re-evaluates the deadline against the live clock and re-arms
+    /// if the target has not yet been reached. This retargets an in-progress
+    /// wait in both directions across a step.
+    pub async fn sleep(self) {
+        loop {
+            let now = self.clock_now();
```
arihant2math, line 61: `Deadline::sleep()` reads `date()`before `clock_set_generation()`. If there is a settime in that window, the wait computes remaining from the old wall clock, then subscribes starting at the new generation, so it misses the step that should retarget the timeout.

src/clock/realtime.rs
```diff
@@ -23,11 +26,60 @@ pub fn set_date(duration: Duration) {
...
+        // Register before re-checking the generation so a step that races our
+        // poll cannot be missed.
+        let mut waiters = clock_set_waiters().lock_save_irq();
+
+        if *CLOCK_SET_GEN.lock_save_irq() != generation {
+            return Poll::Ready(());
+        }
+
+        if !registered {
+            waiters.register(cx.waker());
```
arihant2math: The token is thrown here. If a wrapping `select!` exits via timer/interrupt, the registration stays in the waiter set until the next realtime clock change.

src/process/threading/futex/futex2.rs
```diff
@@ -0,0 +1,258 @@
...
+/// onto `waiters[1].uaddr`. No value comparison is performed (unlike legacy
+/// `FUTEX_CMP_REQUEUE`). Returns the number woken.
+pub async fn sys_futex_requeue(
+    ctx: &ProcessCtx,
+    uwaiters: TUA<FutexWaitvUser>,
+    flags: u32,
+    nr_wake: i32,
+    nr_requeue: i32,
+) -> Result<usize> {
+    // No syscall-level flags are defined.
```
arihant2math, line 225: Linux implements futex2 requeue differently: (`kernel/futex/syscalls.c`, `SYSCALL_DEFINE4(futex_requeue)`). Userspace can observe a successful requeue where it should get EAGAIN or EINVAL.

arihant2math (changes requested):
Mostly LGTM

src/clock/mod.rs
sixseventysix: yeah, that makes sense. I can take that up if you like.

arihant2math: Sure.

arihant2math (approved):
LGTM

